PR #10, opened 2019-03-05
add empty test targets

johnsonj:
- Add a CI script for prow presubmits
- Add doc.go files to packages that have no go files so go test will
  traverse them.


justinsb:
Yay tests!

/approve
/lgtm

PR #11, opened 2019-03-06
Fix merge conflict section in walkthrough

atoato88:
There is a merge conflict at section of "Adding the framework into our types" in docs/addon/walkthrough/README.md
This PR fix conflict by following current contents of pkg/apis/addons/v1alpha1/dashboard_types.go

### DIFF
--- a/docs/addon/walkthrough/README.md
+++ b/docs/addon/walkthrough/README.md
@@ -80,19 +80,15 @@ support (and we can hopefully evolve the contract with just a recompile!)
 
 So in `pkg/apis/addons/v1alpha1/dashboard_types.go`:
 
-<<<<<<< HEAD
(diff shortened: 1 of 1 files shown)


johnsonj:
Thank you!

/assign @justinsb

justinsb:
Awesome - thanks for fixing @atoato88 

/ok-to-test
/approve
/lgtm

johnsonj:
@BenTheElder - is there something else we need to do here to kick off Prow?

BenTheElder:
ah we need to enable the trigger plugin https://prow.k8s.io/plugins?repo=kubernetes-sigs%2Fkubebuilder-declarative-pattern

johnsonj:
This is a real failure (but in HEAD, not from this change)

johnsonj:
Thanks for the fix and patience @atoato88

PR #12, opened 2019-03-06
check in vendor

johnsonj:
I started to go down the path of running dep ensure in CI, but it's a bit problematic. In the future we will switch to go modules and not need the vendor/ folder.


justinsb:
Yes ... bring on go modules (particularly as we are a library), but in the meantime...

/approve
/lgtm

PR #14, opened 2019-03-08
declarative: Add WithApplyPrune

### LINKED ISSUE #4: New declarative option: WithApplyPrune
Allow a declarative operator to use 'kubectl apply --prune' semantics with a provided label.

johnsonj:
This option makes it easier to remove dropped objects inbetween releases
(eg, foo deployment in v1, foo statefulset in v2)

fixes #4

### DIFF
--- a/pkg/patterns/declarative/reconciler.go
+++ b/pkg/patterns/declarative/reconciler.go
@@ -3,6 +3,7 @@ package declarative
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


pkg/patterns/declarative/reconciler.go
```diff
@@ -255,6 +231,21 @@ func (r *Reconciler) applyOptions(opts ...reconcilerOption) {
 	r.options = params
 }
 
+// Validate compatibility of selected options
+func (r *Reconciler) validateOptions() error {
+	var errs []string
+
+	if r.options.prune && r.options.labelMaker == nil {
+		errs = append(errs, "WithApplyPrune must be used with the WithLabels option")
```
justinsb: I had an earlier comment that we should error / warn if prune is used without labels ... but you covered it :thumbsup:

pkg/patterns/declarative/reconciler.go
```diff
@@ -138,19 +118,15 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
-				}
-
-				extraArgs = append(extraArgs, "--prune", "--selector", selectorArg)
-			}
-	*/
+
+	if r.options.prune {
+		var labels []string
+		for k, v := range r.options.labelMaker(ctx, instance) {
+			labels = append(labels, fmt.Sprintf("%s=%s", k, v))
```
justinsb: Things will go wrong if people choose obnoxious labels (e.g. with a comma in them) but I think that's fine - these labels should be drawn either from the CR instance or hard-coded by the operator.

justinsb:
Looks great - I had a concern about validation, but you covered it later in the PR

/approve
/lgtm

PR #16, opened 2019-03-08
smoketest standup

### LINKED ISSUE #6: Basic Smoke Tests
- Bring up a k8s cluster
- Deploy some operators
- Go through the operators lifecycle:
   - create CR, ensure all objects exist
   - delete an object, make sure it gets recreated
   - delete the CR, ensure all objects go away
   - repeat once or twice

johnsonj:
Add a `smoketest.go` to drive the `dashboard-operator` through a few scenarios (create, update, disrupt, destroy).

This validation is scoped to validating framework changes. It may be desirable to make the smoke test driver code part of the library so operator authors can spin up their own.

fixes #6


examples/dashboard-operator/pkg/apis/apis.go
```diff
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 // Generate deepcopy for apis
-//go:generate go run ../../../../vendor/k8s.io/code-generator/cmd/deepcopy-gen/main.go -O zz_generated.deepcopy -i ./... -h ../../hack/boilerplate.go.txt
+//go:generate go run ../../vendor/k8s.io/code-generator/cmd/deepcopy-gen/main.go -O zz_generated.deepcopy -i ./... -h ../../hack/boilerplate.go.txt
```
justinsb: FYI The same trick works here (remove main,go & the path)

justinsb:
Some tweaks are possible, but we can look at those in follow-up.  We should figure out whether we want to use (1) bazel and (2) go modules.

/approve
/lgtm

PR #18, opened 2019-03-08
Implement Application management

johnsonj:
The Application is a CR that represents a given deployment. If a
deployment contains an Application we can specialize it based on the
declarative configuration. We can go further with Addons which have
health/version to expose on the Application.

More info on the application: https://github.com/kubernetes-sigs/application

fixes #5

PR #19, opened 2019-03-09
Implement golden file based tests

### LINKED ISSUE #17: Golden file based tests
The ultimate goal of an operator is to generate and apply a set of YAML manifests to the cluster. Authors should have high confidence that changes to their operator does not unintentionally modify the output.

johnsonj:
This change provides a validator that exercises a given input (CR) to a
declarative.Reconciler matches the expected output (manifest).

fixes #17


johnsonj:
This PR has a dependency on #18

pkg/test/golden/validator.go
```diff
@@ -0,0 +1,329 @@
...
+	}
+
+	actualTmp, err := writeTmp(actual)
+	if err != nil {
+		return errors.Wrapf(err, "write actual yaml to temp file failed")
+	}
+	t.Logf("Wrote actual to %s", actualTmp)
+
+	// pls to use unified diffs, kthxbai?
+	cmd := exec.Command("diff", "-u", expectedPath, actualTmp)
```
justinsb: I think k8s has a diff package, but I think it's a great idea to rely on diff instead as the output from the k8s diff is really awkward, and kops wrote our own diff package and it's a massive pain...

justinsb:
Looks great - and super-happy to see the test failing as it shows the tests are working :+1: 

You do need to rebase though

/approve
/lgtm

justinsb:
Thanks for rebasing!

/approve
/lgtm

PR #20, opened 2019-03-09
declarative/watch: resume watch on failure

### LINKED ISSUE #7: Resume closed watches in DynamicWatch
A Watch will expire after an hour or so (or sooner if the API server restarts). The DyanmicWatch code does not handle resuming the watch.

johnsonj:
A Watch will be dropped by the server periodically. We want to maintain
the watch as long as the operator is running. This change resumes the
Watch then it is closed by the server and maintains the ResourceVersion
to reduce noise.

fixes #7

### DIFF
--- a/pkg/patterns/declarative/pkg/watch/dynamic.go
+++ b/pkg/patterns/declarative/pkg/watch/dynamic.go
@@ -18,17 +18,20 @@ package watch
 
 import (
 	"fmt"
+	"time"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
+
+// WatchDelay is the time between a Watch being dropped and attempting to resume it
+const WatchDelay = 30 * time.Second
 
 func NewDynamicWatch(config rest.Config) (*dynamicWatch, chan event.GenericEvent, error) {
 	dw := &dynamicWatch{events: make(chan event.GenericEvent)}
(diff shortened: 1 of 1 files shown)


pkg/patterns/declarative/pkg/watch/dynamic.go
```diff
@@ -50,32 +54,51 @@ func (dw *dynamicWatch) newDynamicClient(gvk schema.GroupVersionKind) (dynamic.R
...
+	log.WithValues("kind", trigger.String()).WithValues("namespace", target.Namespace).WithValues("labels", options.LabelSelector).WithValues("lastResourceVersion", lastResourceVersion).Info("watch began")
+	for clientEvent := range events.ResultChan() {
+		version, err := accessor.ResourceVersion(clientEvent.Object)
+		if err == nil {
+			lastResourceVersion = version
 		}
-	}()
 
-	return nil
+		if clientEvent.Object == nil || clientEvent.Type == watch.Added {
```
justinsb: It's a little surprising that we always drop watch.Added.  Should we send it and let the client filter?

pkg/patterns/declarative/pkg/watch/dynamic.go
```diff
@@ -50,32 +54,51 @@ func (dw *dynamicWatch) newDynamicClient(gvk schema.GroupVersionKind) (dynamic.R
...
+		}
+	}()
+
+	return nil
+}
+
+func (dw *dynamicWatch) watchUntilClosed(lastResourceVersion string, client dynamic.ResourceInterface, trigger schema.GroupVersionKind, options metav1.ListOptions, target metav1.ObjectMeta) string {
+	log := log.Log
+
+	options.ResourceVersion = lastResourceVersion
```
justinsb: I'm wondering about what happens if lastResourceVersion is too old...

I was thinking there must be a function in our libraries for it - and I find this...
https://github.com/kubernetes/client-go/blob/bad694ec20e3ad1bf3ede898519b4948962639b7/tools/watch/retrywatcher.go

But it looks like that one doesn't handle a "resource version too old" error either... @liggitt should we worry about this scenario?

liggitt: that one explicitly doesn't handle a resourceversiontoo old scenario because it has no way to re-establish a list:

https://github.com/kubernetes/client-go/blob/bad694ec20e3ad1bf3ede898519b4948962639b7/tools/watch/retrywatcher.go#L48-L49

there is a helper that does handle that now: https://github.com/kubernetes/client-go/blob/bad694ec20e3ad1bf3ede898519b4948962639b7/tools/watch/informerwatcher.go#L59-L62

pkg/patterns/declarative/pkg/watch/dynamic.go
johnsonj: agreed! dropped.

pkg/patterns/declarative/pkg/watch/dynamic.go
johnsonj: Thank you! 

It seems that we're better off not managing `ResourceVersion` for now. If we just resume the watch on a failure or drop then we will always reconcile when we need too. We will over-reconcile.

johnsonj: Thank you!

It seems that we're better off not managing `ResourceVersion` for now. If we just resume the watch on a failure or drop then we will always reconcile when we need too. We will over-reconcile.

 WDYT @liggitt ?

liggitt: > If we just resume the watch on a failure or drop then we will always reconcile when we need too. We will over-reconcile.

not having looked through the code, what method would you use to resume the watch without managing a resource version?

johnsonj: It's not really resuming now, it's just restarting with an empty resource version, basically `client.Watch(metav1.ListOptions{})`

johnsonj: I got the idea from your comment here: https://github.com/kubernetes/kubernetes/issues/54878#issuecomment-357575276

liggitt: it is more typical to use an informer in cases like this

justinsb: We did think the optimization here could be to only reset resourceVersion when we observe the "too old" message.  I'm guessing that looks like this: https://github.com/kubernetes/client-go/blob/191b4b1228013836d532375e8cf85986be65edfb/tools/watch/retrywatcher.go#L204

These are going to be watches of the handful of objects each addon-operator created, hopefully tightly filtered by labels.  The reconciliation might be relatively expensive however because we don't (yet) have a way to skip doing an apply - we have to run it speculatively.

I don't love that an informer would buffer the objects in memory, but there won't be many of them, and it could save us from a spurious apply.

And I'm guessing we do now have an example of dynamic informers:

https://github.com/kubernetes/client-go/blob/master/dynamic/dynamicinformer/informer_test.go#L128

justinsb:
OK I think we have a path forwards (thanks @liggitt for clarifying everything) but I think they can be done in follow-on PRs - we want to get (any form of) resume working first!

/approve
/lgtm

pkg/patterns/declarative/pkg/watch/dynamic.go
johnsonj: I filed #39 to better handle this

PR #22, opened 2019-03-10
pkg/patterns/addon/init: Drop flag parsing

johnsonj:
This boilerplate is now included at the top of kubebuilder's generated
main() function. Calling it before the generated code results in
main() declared flags from not being recognized.


justinsb:
Yay progress!

/approve
/lgtm

Issue #3, opened 2019-03-04
Work in Progress


johnsonj:
This mega issue is for capturing the work needed to be able to create an operator with this toolkit.

EXAMPLES/DOCS:
- [x] [Walkthrough](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/tree/master/docs/addon/walkthrough/) is fully functional

TEST INFRA:
- [x] Prow running `hack/ci/test.sh`
- [ ] Prow running `hack/ci/e2e.sh`

PR #23, opened 2019-03-10
docs/addon/walkthrough: Cleanup pass

johnsonj:
- Remove bazel for direct dockerfile builds.
- Reference actual repo path and don't rely on unimplemented kubebuilder
  flag.
- Rely on kubebuilders RBAC generation and give the set of roles needed
  for the dashboard operator
- Tighten up some language

updates #3

### DIFF
--- a/docs/addon/walkthrough/README.md
+++ b/docs/addon/walkthrough/README.md
@@ -1,24 +1,37 @@
-## Walkthrough of creating a new operator
+## Walkthrough: Creating a new Operator
 
-This walkthrough is for creating a new operator, to run the kubernetes
-dashboard.
+This walkthrough is for creating an operator to run the kubernetes dashboard.
 
 ### Basics
 
-Install kubebuilder from https://github.com/kubernetes-sigs/kubebuilder/releases if you have not yet already done so.
+Install the following depenencies:
 
-Create a new directory and use kubebuilder to create a basic operator for you:
+- [kubebuilder](https://book.kubebuilder.io/getting_started/installation_and_setup.html)
+- [kustomize](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/INSTALL.md)
+- docker
+- kubectl
+- golang
+- [dep](https://github.com/golang/dep#installation)
+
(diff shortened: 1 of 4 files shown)


docs/addon/walkthrough/README.md
```diff
@@ -240,48 +286,106 @@ should see the deployment etc that the operator has created.
...
-)
-EOF
-```
+1. Modify the memory limit for the Operator located in: `config/manager/manager.yaml`:
+
+	```yaml
+	apiVersion: apps/v1
+	kind: StatefulSet
+	spec:
+	  template:
```
atoato88:
```
template:
   spec: # <- it's three space before `spec`. should be two space?
```

docs/addon/walkthrough/README.md
```diff
@@ -240,48 +286,106 @@ should see the deployment etc that the operator has created.
...
+	     spec:
```
atoato88: Is it missing `containers:` ?
What do you think about below?
```
apiVersion: apps/v1
kind: StatefulSet
spec:
  template:
    spec:
      containers:
      - name: manager
        resources:
          limits:
            cpu: 100m    # unchanged
            memory: 150m # updated
```

justinsb: Or maybe just drop everything but the `resources` block?  Unless you intended it to be a kustomize patch?

justinsb:
Looks good - thanks for doing this @johnsonj 

Maybe tweak the resources yaml, and I'd personally just use `maintainer@example.org` instead of being under pressure to keep making up fun names...

docs/addon/walkthrough/README.md
johnsonj: Thanks to you both for looking a this! `kustomize` patch is a great idea. I'll do that.

docs/addon/walkthrough/README.md
johnsonj: ACK

justinsb:
Thanks @johnsonj 

/approve
/lgtm

PR #27, opened 2019-03-19
dashboard-operator: default image name to $USER/dashboard-operator

johnsonj:
This is a more reasonable default for those lucky enough that `$USER == $dockerhub login`

### DIFF
--- a/examples/dashboard-operator/config/default/manager_image_patch.yaml
+++ b/examples/dashboard-operator/config/default/manager_image_patch.yaml
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: gcr.io/jrjohnson-gke/dashboard-operator:latest
+      - image: /dashboard-operator:latest
         name: manager
--- a/examples/dashboard-operator/Makefile
+++ b/examples/dashboard-operator/Makefile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Image URL to use all building/pushing image targets
-IMG ?= gcr.io/jrjohnson-gke/dashboard-operator:latest
+IMG ?= ${USER}/dashboard-operator:latest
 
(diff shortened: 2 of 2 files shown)


examples/dashboard-operator/config/default/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: gcr.io/jrjohnson-gke/dashboard-operator:latest
+      - image: /dashboard-operator:latest
```
justinsb: This is a tricky problem - I agree we should merge as-is until we find a better approach!

justinsb:
We need to figure out how to link up the user variable substitution into kubebuilder, but in the meantime

/approve
/lgtm

examples/dashboard-operator/config/default/manager_image_patch.yaml
atoato88: If `kustomize build` generate manifest from this, `image` is set to `/dashboard-operator:latest` without above comment( it say "Change the value of image ..." ).
IMHO, it is better that add some prefix saying "change me" as value of 'image'.
e.g. `<my-cool-project>/dashboard-operator:latest` or `<change-me>/dashboard-operator:latest` or something else.

atoato88: Oh, it's merged, sorry for late comment.

johnsonj: This should be filled in by the kubebuilder generated: https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/examples/dashboard-operator/Makefile#L69 and ultimately consumed by kustomize. 

I agree an invalid image name here would be a better default than just the `/..`

PR #28, opened 2019-03-27
Expose NewObject function in manifest package externally

kawych:


pkg/patterns/declarative/pkg/manifest/objects.go
```diff
@@ -337,7 +337,7 @@ func ParseObjects(ctx context.Context, manifest string) (*Objects, error) {
 	return objects, nil
 }
 
-func newObject(u *unstructured.Unstructured, json []byte) (*Object, error) {
+func NewObject(u *unstructured.Unstructured, json []byte) (*Object, error) {
```
justinsb: So it's possible to misuse this method now - setting json inconsistent with unstructured.Unstructured.

If you're starting with json, the safer method is ParseJSONToObject

If you're starting with unstructured.Unstructured, we should expose `NewObject` that doesn't allow you to set `json`

(`json` acts as a cache of the generated json anyway - it's just an optimization)

Looks like we also need some docs to steer callers to the right methods here.

What's your use-case though?  Starting from json / starting from unstructured.Unstructured?

kawych: My use-case is starting from unstructured. I've modified it to expose NewObject without json argument.

justinsb:
Thanks @kawych - makes a ton of sense, and thanks for making the change!

/approve
/lgtm

PR #29, opened 2019-03-28
Make the default manifest loader pluggable

justinsb:
This enables it to be configured or changed.

### DIFF
--- a/pkg/patterns/declarative/options.go
+++ b/pkg/patterns/declarative/options.go
@@ -24,8 +24,17 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/addon/pkg/loaders"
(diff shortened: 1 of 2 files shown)


pkg/patterns/declarative/options.go
```diff
@@ -24,9 +24,18 @@ import (
...
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/addon/pkg/loaders"
 	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative/pkg/manifest"
 )
 
+type ManifestLoaderFunc func() ManifestController
+
+// DefaultManifestLoader is the manifest loader we use when a manifest loader is not otherwise configured
+// We curently default to loading from the filesystem, but may change this default in future
+var DefaultManifestLoader ManifestLoaderFunc = func() ManifestController {
```
johnsonj: Thanks for adding this plug-ability. I agree we want a default at the 'binary level'. 

I wonder if (in the future) we can get away with not specifying the default here to have a proper separation of declarative/addon. 

We could move this code into addon.Init() for example.

johnsonj:
Looks good, no reason to hold back based on my comment as this change does not introduce the dependency, just moves it. I'll do a follow up to set the default in addon.Init() if that seems reasonable.

/lgtm
/approve

PR #32, opened 2019-03-28
declarative,addon: Configure the filesystem loader in addon.Init

johnsonj:
This change removes the dependency between the declarative and addon
package. This allows non-addon users to utilize the declarative
reconciler with their own manifest loader.


justinsb:
Ah - very nice!  This dependency always bugged me a little!

/approve
/lgtm

PR #37, opened 2019-04-02
declarative,docs: remove packageName

### LINKED ISSUE #33: packageName is ignored by filesystem loader
[declarative/reconciler.Init](https://godoc.org/sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative#Reconciler.Init) requires a `packageName` parameter. This value is never propagated to [ManifestController.ResolveManifest](https://godoc.org/sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative#ManifestController). The addon filesystem loader relies on [CommonObject.ComponentName](https://godoc.org/sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/addon/pkg/apis/v1alpha1#CommonObject).

We should have just one of these, likely packageName.

johnsonj:
The addon manifest loader uses the 'ComponentName()' call on
'CommonSpec' to resolve the component to load the manifest from.

We could do this directly in declarative by requiring packageName and
bubbling it out to the Loader, but there's no reason for the declarative
controller to keep track of this value. The library can't make decisions
about which package to use beyond passing a string it was given.

Pushing this out to the addon/loader allows implementers to own the
decision and perhaps change it at runtime based on the CRs configuration
or the state of CLI flags.

fixes #33

### DIFF
--- a/docs/addon/walkthrough/README.md
+++ b/docs/addon/walkthrough/README.md
@@ -188,7 +188,7 @@ func newReconciler(mgr manager.Manager) *ReconcileDashboard {
 
 	r := &ReconcileDashboard{}
 
-	r.Reconciler.Init(mgr, &api.Dashboard{}, "dashboard",
+	r.Reconciler.Init(mgr, &api.Dashboard{},
 		declarative.WithObjectTransform(declarative.AddLabels(labels)),
 		declarative.WithOwner(declarative.SourceAsOwner),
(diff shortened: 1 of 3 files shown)


justinsb:
Ah - so we weren't even using packageName?  In that case, nice cleanup - thanks!

/approve
/lgtm

PR #38, opened 2019-04-02
declarative,addon: global pipeline, basic registry support

### LINKED ISSUE #35: Support basic private registry for docker environments
Provide a set of flags for addon operators that, when specified, rewrite the image names and add add a imagePullSecret to relevant PodSpecs.

Flags:
- `--private-registry=<my-repo>`
- `--image-pull-secret=<my-secret>`

johnsonj:
- Add the Options.{Begin,Eng} variable to declarative which enables users
  to specify global transforms for the Reconciler.
- Add --{private-registry,image-pull-secret} flags to addon operators and
  inject the ImageRegistryTransform when set.

fixes #35

### DIFF
--- a/examples/dashboard-operator/cmd/manager/main.go
+++ b/examples/dashboard-operator/cmd/manager/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"flag"
 	"os"
 
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
@@ -32,6 +33,7 @@ import (
 
 func main() {
 	addon.Init()
+	flag.Parse()
 
 	logf.SetLogger(logf.ZapLogger(false))
 	log := logf.Log.WithName("entrypoint")
--- a/pkg/patterns/declarative/image.go
+++ b/pkg/patterns/declarative/image.go
@@ -1,0 +1,82 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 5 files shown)


examples/dashboard-operator/cmd/manager/main.go
```diff
@@ -32,6 +33,7 @@ import (
 
 func main() {
 	addon.Init()
```
justinsb: Wondering if we should put flag.Parse() into addon.Init (and require that you declare flags before calling addon.Init.  Or alternatively we could require that you have called flag.Parse _before_ calling addon.Init - that might be easier to enforce.  But this shouldn't block this PR

pkg/patterns/declarative/image.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func applyImageRegistry(ctx context.Context, operatorObject DeclarativeObject, manifest *manifest.Objects, registry, secret string) error {
+	if registry == "" && secret == "" {
+		return nil
+	}
+	for _, manifestItem := range manifest.Items {
+		if manifestItem.Kind == "Deployment" || manifestItem.Kind == "DaemonSet" ||
+			manifestItem.Kind == "StatefulSet" || manifestItem.Kind == "Job" ||
+			manifestItem.Kind == "CronJob" {
```
justinsb: We might later want to extract this into a "visit images" helper.  But only when we need it next!

pkg/patterns/declarative/options.go
```diff
@@ -32,6 +29,14 @@ type ManifestLoaderFunc func() ManifestController
 // DefaultManifestLoader is the manifest loader we use when a manifest loader is not otherwise configured
 var DefaultManifestLoader ManifestLoaderFunc
 
+// Options are a set of reconcilerOptions applied to all controllers
+var Options struct {
+	// Begin options are applied before evaluating controller specific options
+	Begin []reconcilerOption
```
justinsb: Pre and Post might be more intuitive ... not sure!

justinsb:
Nice improvements!

/approve
/lgtm

PR #44, opened 2019-05-08
Add Patch

johnsonj:
Introduce Patch. An Addon can implement the Patchable interface to
accept a set of Patches that are overlaid on the deployment. For
example:

```yaml
apiVersion: addons.sigs.k8s.io/v1alpha1
kind: Dashboard
metadata:
  name: dashboard-sample
  namespace: kube-system
spec:
  patches:
  - apiVersion: apps/v1beta2
    kind: Deployment
    metadata:
      name: kubernetes-dashboard
      namespace: kube-system
    spec:
      replicas: 5
```

This patch will change the replica count of the dashboard deployment.
This allows us to provide a break-glass experience for the user to set
configuration options while not requiring every configuration option be
directly exposed on the CRD.

- It's not clear if we want this on all addons. If so, we should
  graduate it to CommonObject
- This patching is equivalent/borrowed from kustomize which is not
  currently avaliable as a library.


stealthybox:
Would we want to be able to specify kustomize [bases](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/kustomization.yaml#L143) as well?

This would allow you to check your overlays/patches into source control and use overlays that are distributed via channels or the operator's file system.

johnsonj:
That seems interesting @stealthybox! 

I think we need to think about if the patches on the CRD are the 'final assembly' or the a higher level (declaration of the intent?). If they're the sort of final instructions, then I think we can keep `patch` as simple as possible. If they're higher level then should we just completely accept a kustomization field? 

I guess we should think about: 'would a user patch the CRD with kustomize' or 'is this CRD an interface into kustomize'?

examples/dashboard-operator/pkg/controller/dashboard/dashboard_controller.go
```diff
@@ -71,8 +71,9 @@ func newReconciler(mgr manager.Manager) (r *ReconcileDashboard, srcLabels declar
 		declarative.WithStatus(status.NewBasic(mgr.GetClient())),
 		declarative.WithPreserveNamespace(),
 		declarative.WithApplyPrune(),
-		declarative.WithManagedApplication(srcLabels),
 		declarative.WithObjectTransform(addon.TransformApplicationFromStatus),
+		declarative.WithManagedApplication(srcLabels),
+		declarative.WithObjectTransform(addon.ApplyPatches),
```
justinsb: It is going to be weird if the field is there, but the patches don't actually do anything.  Maybe this should be the default?

pkg/patterns/declarative/application.go
```diff
@@ -38,8 +38,19 @@ func transformApplication(ctx context.Context, instance DeclarativeObject, objec
...
+		convertedLabels[k] = v
+	}
+	labelSelector := map[string]interface{}{"matchLabels": convertedLabels}
+	app.SetNestedField(labelSelector, "spec", "selector")
+
+	componentGroupKinds := []interface{}{}
+	for _, gk := range uniqueGroupKind(objects) {
+		componentGroupKinds = append(componentGroupKinds, map[string]interface{}{"group": gk.Group, "kind": gk.Kind})
+	}
+	app.SetNestedSlice(componentGroupKinds, "spec", "componentGroupKinds")
```
justinsb: Looks good, though why are we doing this?  Just to break the dependency on metav1?

justinsb:
This lgtm.

On the base question, I think we can consider that separately from this PR @stealthybox ?  My guess is that the base should be defined by the operator itself - or maybe by a version/channel field in the object, and that we wouldn't expect to allow arbitrary bases.  But that's just my guess!

examples/dashboard-operator/pkg/controller/dashboard/dashboard_controller.go
johnsonj: This is necessary today because the addons package defines the types, so the declarative package can't extract the patch out. We _could_ workaround this with reflection.

I think what we really need is an AddonsReconciler that applies all (or most?) of these transforms.

pkg/patterns/declarative/application.go
johnsonj: I had to do this because the actual metav1.LabelSelector{} couldn't be deep copied (which is why SetNestedFieldNoCopy) was used.

justinsb:
Thanks @johnsonj 

/lgtm

johnsonj:
hmm, no failure locally, let's see:


/test pull-declarative-test
--

justinsb:
Thanks for rebasing & fixing tests!

/lgtm

PR #45, opened 2019-05-20
docs/dashboard-operator: switch to example.org

johnsonj:
k8s.io is reserved for Kubernetes APIs. We don't want users to assume
this domain can be used by any extension or addon. The API group
should be a valid domain controlled by the author (or one dedicated to
examples such as example.{com,net,org})

### DIFF
--- a/docs/addon/walkthrough/README.md
+++ b/docs/addon/walkthrough/README.md
@@ -19,7 +19,7 @@ Create a new directory and use kubebuilder to scafold the operator:
 cd $(go env GOPATH)
 mkdir -p src/sigs.k8s.io/dashboard-operator/
 cd src/sigs.k8s.io/dashboard-operator/
-kubebuilder init --domain sigs.k8s.io --license apache2 --owner "The Kubernetes Authors" --dep=true
+kubebuilder init --domain addons.example.org --license apache2 --owner "The Kubernetes Authors" --dep=true
 ```
 
 Add the patterns to your project:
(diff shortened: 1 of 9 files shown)


bgrant0607:
cc @derekwaynecarr

docs/addon/walkthrough/README.md
```diff
@@ -19,7 +19,7 @@ Create a new directory and use kubebuilder to scafold the operator:
 cd $(go env GOPATH)
 mkdir -p src/sigs.k8s.io/dashboard-operator/
 cd src/sigs.k8s.io/dashboard-operator/
-kubebuilder init --domain sigs.k8s.io --license apache2 --owner "The Kubernetes Authors" --dep=true
+kubebuilder init --domain addons.example.org --license apache2 --owner "The Kubernetes Authors" --dep=true
```
bgrant0607: Please change "The Kubernetes Authors" also

bgrant0607:
In general, we should probably standardize on example.com and/or example.org for API groups, label and annotation prefixes, etc. 

https://www.iana.org/domains/reserved

I don't know whether we have a good place to document that. 
cc @liggitt

docs/addon/walkthrough/README.md
johnsonj: done, thanks!

derekwaynecarr:
nice catch, thanks for following up @bgrant0607

docs/addon/walkthrough/README.md
```diff
@@ -19,7 +19,7 @@ Create a new directory and use kubebuilder to scafold the operator:
 cd $(go env GOPATH)
 mkdir -p src/sigs.k8s.io/dashboard-operator/
 cd src/sigs.k8s.io/dashboard-operator/
-kubebuilder init --domain sigs.k8s.io --license apache2 --owner "The Kubernetes Authors" --dep=true
+kubebuilder init --domain addons.example.org --license apache2 --owner "TODO($USER): assign copyright" --dep=true
```
justinsb: Should this just be `example.org` ?

justinsb:
Thanks @johnsonj 

/approve
/lgtm

Issue #47, opened 2019-07-04
Add setter to FakeClient for setting arbitrary value


atoato88:
For `Golden File Tests`, it is useful that `FakeClient` be eble to set specific value to [`c` field](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/pkg/test/mocks/client.go#L16).
Currently, `FakeClient` doesn't have setter for its `c` field.

atoato88:
related: kubernetes-sigs/addon-operators#19

PR #50, opened 2019-07-10
Implement Create method in FakeClient

atoato88:
Currently, `FakeClient` doesn't implement `Create` method.
Some tests need to set specific resource object.
This PR implements `Create` method in `FakeClient`.

Fixes: #47

Related: kubernetes-sigs/addon-operators#19

### DIFF
--- a/pkg/test/mocks/client.go
+++ b/pkg/test/mocks/client.go
@@ -2,30 +2,42 @@ package mocks
 
 import (
 	"context"
-	"fmt"
-	"reflect"
 
+	"encoding/json"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/testing"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 )
 
 // FakeClient is a struct that implements client.Client for use in tests.
 type FakeClient struct {
-	c             map[client.ObjectKey]runtime.Object
 	ErrIfNotFound bool
+	tracker       testing.ObjectTracker
+	scheme        *runtime.Scheme
 }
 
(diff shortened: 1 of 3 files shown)


atoato88:
I wonder that [here](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/50/files#diff-61530f6552ccccc888b17fa3fe9f34abR16) has some ambiguous, because ObjectKey has only two fields which are Namespace and Name.
Some tests may register different Object with same ObjectKey.
When this case, FakeClient returns only one Object.
So, is it better to change `c: map[runtime.ObjectKey]runtime.Object` with `c: map[schema.GroupVersionResource][]runtime.Object` like [`k8s.io/client-go/testing`](https://github.com/kubernetes/client-go/blob/master/testing/fixture.go#L200) or use `testing.ObjectTracker` in `k8s.io/client-go/testing`?

Any recommends or comments? or when it's my misunderstanding, please let me know.

pkg/test/mocks/client.go
```diff
@@ -48,8 +48,14 @@ func (FakeClient) List(ctx context.Context, list runtime.Object, opts ...client.
...
 
-func (FakeClient) Create(ctx context.Context, obj runtime.Object, opts ...client.CreateOptionFunc) error {
-	panic("not implemented")
+func (f FakeClient) Create(ctx context.Context, obj runtime.Object, opts ...client.CreateOptionFunc) error {
+	resource := reflect.ValueOf(obj).Elem()
+	namespace := resource.FieldByName("ObjectMeta").FieldByName("Namespace").String()
+	name := resource.FieldByName("ObjectMeta").FieldByName("Name").String()
+
+	key := client.ObjectKey{Namespace: namespace, Name: name}
+	f.c[key] = obj
```
johnsonj: Maybe error or panic if it already exists?

Though it's a fine line between mock and re-implementing the apiserver :)

johnsonj:
I think you understanding of the issue is spot on. Create(Deployment{Foo, Bar}) would overwrite Create(StatefulSet{Foo, Bar}). I think this hasn't been an issue so far because the only use case of it is to provide a pre-populated set of objects (so it'd be pretty obvious you're overwriting it).

I think either solution is reasonable. At a quick glance, ObjectTracker looks really nice, it seems like if you switched to that we could drop a good deal of code in the FakeClient.

atoato88:
I updated an implementation of FakeClient which use ObjectTracker.

johnsonj:
Thanks @atoato88 !

/lgtm
/approve

PR #51, opened 2019-08-20
Allow watching objects with the same name but different namespaces

Catramen:
by appending namespace to the registered key.


johnsonj:
Thank you @Catramen !

/ok-to-test
/lgtm
/approve

PR #53, opened 2019-09-26
fix bug in less function, it should compare different values

beautytiger:


johnsonj:
Thanks @beautytiger - great find! that must have been tricky.

PR #57, opened 2019-10-17
Hash configMap's name

annp1987:
Hash configMap's name

A hash is appended to configMap's name and any change in the ConfigMap will trigger a rolling update. This allow us to resolve the issue https://github.com/kubernetes-sigs/addon-operators/issues/27

Signed-off-by: Nguyen Phuong An <AnNP@vn.fujitsu.com>


johnsonj:
Thank you for this @annp1987 ! We definitely want kustomize support and this is a great way to get things started.

Could you describe the user workflow for this change? (How could I setup kustomize patches for this?). Does this enable new scenarios that aren't covered in the existing [patch](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/commit/92a107d01b52ed2d3d8f05a835b3f9e4cfbee080)?

It would be great to add an example of how to use this by creating a new [version for the dashboard-operator](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/tree/master/examples/dashboard-operator/channels/packages/dashboard).

If you'd like to talk about this with higher bandwidth, we have a [bi-weekly cluster-addons meeting](https://github.com/kubernetes/community/tree/master/sig-cluster-lifecycle) if the time works for you. Otherwise, I'm Github issues are a great way to discuss it.

annp1987:
Hi @johnsonj,

I appreciated your quickly response. Following your comment at https://github.com/kubernetes-sigs/addon-operators/issues/27#issuecomment-543245766. I would like to introduce ConfigSpec instead of supporting purely 'kustomize build' command. For example:
```yaml
apiVersion: addons.sigs.k8s.io/v1alpha1
kind: Dashboard
metadata:
  name: dashboard-sample
  namespace: kube-system
spec:
  config:
  - apiVersion: apps/v1beta2
    kind: Deployment
    metadata:
      name: kubernetes-dashboard
      namespace: kube-system
    spec:
      data: path-to-config-file
      confighash: true
```
This will replace ConfigMap's data and generate a hash ConfigMap name of the dashboard deployment. So we can do rolling upgrade for dashboard or coredns. I'm not sure my understanding is correct or not. Could you please give me some advice? Thanks in advance. :-)

annp1987:
Hi @johnsonj,

After double thinking, I would like support ConfigMapGenerator for kubebuilder-declarative-pattern. So a CRD of CoreDNS would look like:
``` yaml
apiVersion: addons.k8s.io/v1alpha1
kind: CoreDNS
metadata:
  name: default
  namespace: kube-system
spec:
  channel: stable
  patches:
  …..
  configMapGenerator:
  - name: coredns
   namespace: kube-system
    files:
    - Corefile
```
Does it make sense to you?

annp1987:
> Thank you for this @annp1987 ! We definitely want kustomize support and this is a great way to get things started.
> 
> Could you describe the user workflow for this change? (How could I setup kustomize patches for this?). Does this enable new scenarios that aren't covered in the existing [patch](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/commit/92a107d01b52ed2d3d8f05a835b3f9e4cfbee080)?
> 
> It would be great to add an example of how to use this by creating a new [version for the dashboard-operator](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/tree/master/examples/dashboard-operator/channels/packages/dashboard).
> 
> If you'd like to talk about this with higher bandwidth, we have a [bi-weekly cluster-addons meeting](https://github.com/kubernetes/community/tree/master/sig-cluster-lifecycle) if the time works for you. Otherwise, I'm Github issues are a great way to discuss it.

annp1987:
Hi @johnsonj,

I really appreciated your suggestion at  https://github.com/kubernetes-sigs/addon-operators/issues/27#issuecomment-543245766. So I would like to introduce a hash configMap's name function to the framework. Hash function is borrowed from kustomize project. I'm not sure my understanding your suggestion is correct or not. Could you please teach me? Thanks in advance. :-)

annp1987:
Hi Jeff,

Should we integrate kustomize with the framework via krusty package as short term solution? It means that  MakeKustomizer will be called at https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/855438fd2481bc911801257efba1e23bda6f0b8a/pkg/patterns/addon/pkg/loaders/types.go#L125, right?

johnsonj:
Hey @annp1987 - Yes to the krusty package.

I don't think that's the right spot. This is too early to invoke kustomize. We really want to do all of our transformations _then_ run kustomize to finalize the manifest.

One way to enable this:
- Adapt the LoadManifest call to load multiple files (everything in the directory). The loader should annotate the file path for each file. Perhaps instead of LoadManifest returning string, it returns map[string]string (where the key is the file path). [ManifestLoader.ResolveManifest](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/855438fd2481bc911801257efba1e23bda6f0b8a/pkg/patterns/addon/pkg/loaders/fs.go#L49) should be updated to return this as well.
- Update [Reconciler.BuildDeploymentObjects](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/855438fd2481bc911801257efba1e23bda6f0b8a/pkg/patterns/declarative/reconciler.go#L169) to handle the new format. The most straightforward way to do this would be to loop through each file and run all of the transforms against it.
- Now we can call NewKustomizer and run [kustomize.Run](https://github.com/kubernetes-sigs/kustomize/blob/api/v0.2.0/api/krusty/kustomizer.go#L47) in Reconciler.BuildDeploymentObjects (or after the call to [injectOwnerRef](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/855438fd2481bc911801257efba1e23bda6f0b8a/pkg/patterns/declarative/reconciler.go#L123)). To do this, you will either need to write out all of the files to a temporary directory, or implement [filesys.FileSystem](https://github.com/kubernetes-sigs/kustomize/blob/api/v0.2.0/api/filesys/filesystem.go#L12) that uses the objects as they exist in memory. You may be able to utilize (or just use) the [fsInMemory](https://github.com/kubernetes-sigs/kustomize/blob/api/v0.2.0/api/filesys/fsinmemory.go) for this.

PR #58, opened 2019-10-21
controller-runtime: pkg/runtime/log -> pkg/log

justinsb:
pkg/runtime/log is now deprecated (and pulls in the zap library, which
shouldn't be required)


johnsonj:
/lgtm

Thanks @justinsb !

PR #59, opened 2019-10-21
Update go.mod dependency to kubebuilder 0.2.2

justinsb:
We shouldn't be pinnned to a beta version, now that a released version
of 0.2 is available.


johnsonj:
Nice! great promotion :)

/lgtm

PR #61, opened 2019-10-21
Replace yaml.v2 with sigs.k8s.io/yaml

justinsb:
We should use the k8s version of the libraries.


atoato88:
This PR is OK to me, because it has no errors on my lab environment.

PR #62, opened 2019-10-21
Support multiple packages in a channel manifest file

justinsb:
We now allow a package name to be specified, but manifests with no
name are assumed to match all packages.  Where there is a tie, we
prefer a named package, and then we prefer the newer version.

(Previous version of this commit was a breaking change ... no longer)

### DIFF
--- a/pkg/patterns/addon/pkg/loaders/types.go
+++ b/pkg/patterns/addon/pkg/loaders/types.go
@@ -23,7 +23,9 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/blang/semver"
 	yaml "gopkg.in/yaml.v2"
+	"k8s.io/klog"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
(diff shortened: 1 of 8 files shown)


johnsonj:
Thanks @justinsb ! This has a failing test (fail to build), it might need rebased on master w/ the latest version bumps?

pkg/patterns/addon/pkg/loaders/types.go
```diff
@@ -132,21 +132,25 @@ func (r *FSRepository) LoadManifest(ctx context.Context, packageName string, id
...
+	for i := range c.Packages {
+		v := &c.Packages[i]
+		if v.Package != "" && v.Package != packageName {
+			continue
+		}
 		if latest == nil {
 			latest = v
-		} else {
-			return nil, fmt.Errorf("version selection not implemented")
+		} else if latest.Compare(v) < 0 {
```
atoato88: In my lab environment, Error occurs following.
Missing definition of function?
```
latest.Compare undefined (type *Version has no field or method Compare)
```

pkg/patterns/addon/pkg/loaders/types.go
```diff
@@ -132,21 +132,25 @@ func (r *FSRepository) LoadManifest(ctx context.Context, packageName string, id
...
+			latest = v
 		}
```
atoato88: Isn't `else {...}` needed?
If both `if{...}` and `else if ... {...}` aren't match its state, this function returns `latest` as nil. Or is it ok?

pkg/patterns/addon/pkg/loaders/types.go
justinsb: Thanks - I had mixed up my branches!

pkg/patterns/addon/pkg/loaders/types.go
justinsb: The missing if branch is when we already have a newer version of the package.

e.g.

```
packages:
- name: foo
  version: 2
- name: foo
  version: 1
```

latest will be set to the the foo v2 on the first iteration, and then we ignore foo v1 because it is older.

I'm not sure how useful it is to actually have multiple matching packages, but the scenario I wanted to enable was this:

```
packages:
- version: 2.0.0 # default version
- name: foo
  version: 1.0.0 # foo is pinned to a particular version
```

atoato88: I see. :smiley: 
I recognize that I have misunderstanded a logic..
Thanks to your kind explanation.

pkg/patterns/addon/pkg/loaders/types.go
atoato88: I've checked it. :+1:

pkg/patterns/addon/pkg/loaders/versions_test.go
```diff
@@ -0,0 +1,91 @@
...
+		{
+			Bigger:  Version{Package: "", Version: "2.0.0"},
+			Smaller: Version{Package: "", Version: "1.0.0"},
+		},
+		// Parsing is tolerant
+		{
+			Bigger:  Version{Package: "", Version: "1"},
+			Smaller: Version{Package: "", Version: "0.1"},
+		},
+		// Named packages have higher priority
```
johnsonj: this is a sane default/"upgrade"- thanks!

PR #63, opened 2019-10-21
correct sample code

vincent-pli:
Remove the line to avoid exception `imported and not used: addon`


johnsonj:
Thank you @vincent-pli !

/ok-to-test
/lgtm
/approve

PR #67, opened 2020-01-07
Update to controller-runtime 0.4.0

justinsb:


johnsonj:
@justinsb - Looks like a small API change is needed:

(from the tests):
```
/home/prow/go/pkg/mod/k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible/rest/request.go:598:31: not enough arguments in call to watch.NewStreamWatcher
	have (*versioned.Decoder)
	want (watch.Decoder, watch.Reporter)
FAIL	dashboard-operator/controllers [build failed]
```

justinsb:
Thanks @johnsonj - a go.mod bump it looks like, but thanks for pointing it out!  I'll probably have to update #68 in the same way :-)

PR #69, opened 2020-01-11
Change imports in examples directory to use fully qualified path

davis-haba:
Installing kubebuilder-declarative-pattern with "go get sigs.k8s.io/kubebuilder-declarative-pattern/..." fails because of a few relative import paths in the examples directory (the '...' in the github URL causes 'go get' to recursively install subdirectories). Changing these to use the fully qualified paths should fix this.


johnsonj:
Hey @davis-haba is your version of golang using gomodules? I thought it 'stops' once it gets to a go.mod file. If memory serves, I used the short paths to workaround some stuff.. but I can't quite recall what they were/if they are necessary.

davis-haba:
Hey @johnsonj ,

Thanks for taking a look this PR :)

My go version is indeed using Go modules (go version go1.13.4 linux/amd64). 

I'm not entirely sure how `go get` works with respect to go.mod files in sub directories, but what I've observed is that while `go get sigs.k8s.io/kubebuilder-declarative/pattern` works just fine, the inclusion of the ellipse at the end of the path (i.e. `go get sigs.k8s.io/kubebuilder-declarative/pattern/...`) does not work without this proposed fix. The "..." causes `go get` to fetch the package's sub directories/dependencies as well, which fails when sees the relative import paths in a few files in the examples directory and tries to install them. 

Here is the specific error I see from running `go get sigs.k8s.io/kubebuilder-declarative/pattern/...`:
```
package sigs.k8s.io/kubebuilder-declarative-pattern/examples/dashboard-operator
	imports dashboard-operator/api/v1alpha1: unrecognized import path "dashboard-operator/api/v1alpha1" (import path does not begin with hostname)
package sigs.k8s.io/kubebuilder-declarative-pattern/examples/dashboard-operator
	imports dashboard-operator/controllers: unrecognized import path "dashboard-operator/controllers" (import path does not begin with hostname)
```

I tested out the changes on my fork (changing sigs.k8s.io to github.com/davis-haba where appropriate), and it appears to fix the issue. 

A tool I am using is forcing me to use this ellipse path URL in `go get` which is why I'm trying to fix this upstream.

johnsonj:
Thanks @davis-haba ! 

/lgtm

PR #72, opened 2020-01-27
Fix command for generate patch yaml file

erain:
Fixed the command for generate `manager_resource_patch.yaml` file: `echo` -> `cat`

Also removed all trailing spaces.


johnsonj:
thanks @erain !

/ok-to-test
/lgtm
/approve

PR #73, opened 2020-02-05
Add metrics-server-operator

atoato88:
This PR adds metrics-server-operator to `examples/`.
As of now dashboard-operator is located in `examples/`.
[Dashboard](https://github.com/kubernetes/dashboard/) v2.0.0 which is next release can use [metrics-server](https://github.com/kubernetes-sigs/metrics-server) for
monitoring statistics of k8s nodes.

It's better to add metrics-server-operator to this repo because that
dashboard v2.0.0 will release before long.

When dashboard v2.0.0 has released, it is needed to add v2.0.0
manifests to dashboard-operator.


atoato88:
/assign @johnsonj
Could you review this?

examples/metrics-server-operator/api/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
+/*
+Copyright 2020 TODO(akihito): assign copyright.
```
johnsonj: I think The Kubernetes Authors is standard?

johnsonj:
Thank you @atoato88 !

examples/metrics-server-operator/api/v1alpha1/groupversion_info.go
atoato88: Thank you for comment. I've fixed these.

atoato88:
@johnsonj 
Thank you for approving!!

I also thought what you considered. :smiley:
dashboard-operator and metrics-server-operator in this repository is examples for this framework.
But the same time, it is reasonable to place in kubernetes-sigs/cluster-addons because of it's truely addons. :smiley:

As you comment, we can move these to kubernetes-sigs/cluster-addons.
But I think that it should not be duplicated to two places(kubebuilder-declarative-patterns/examples/ and cluster-addons/) becuase of avoiding double management.
WDYT?

Should we use `git submodule`?

1. move codes(relating to dashboard-operator and metrics-server-operator) in kubebuilder-declarative-patterns/examples to cluster-addons repository
1. remove directory `examples` on kubebuilder-declarative-patterns
1. on kubebuilder-declarative-patterns, `git submodule add https://github.com/kubernetes-sigs/cluster-addons.git examples`

But this may need some code changes `import` in *.go files and go.mod.

Or, is `git submodule` something tricky?

johnsonj:
I think you nailed it that dashboard is an addon and duplication is bad news. Submodules are tough to manage. If we want to make the dashboard-operator useful (and not just an illustrative example of how to use this library) then I think it just doesn't belong here. 

What about a wilder turn: 
1. Move `{dashboard, metrics-server}-operator` to cluster-addons
1. Replace `dashbboard-operator` with something where we don't care about versions, like `guestbook-operator` ([as in this guestbook](https://github.com/kubernetes/examples/tree/master/guestbook) or the go version). We can just use it to show 'how it's made'

atoato88:
I agree with your idea. :+1: 
It's a reasonable suggestion.

PR #81, opened 2020-04-07
Remove dashboard-operator from examples

atoato88:
dashboard-operator moved to `sigs.k8s.io/cluster-addons` by [PR on cluster-addons repository](https://github.com/kubernetes-sigs/cluster-addons/pull/44)
This PR removes dashboard-operator from `exmaples` folder on this repository.
`hack/ci/test.sh` is also changed to replace dashboard-operator with
guestbook-operator.

related to: #80

Issue #80, opened 2020-04-06
Remove dashboard-operator from examples


atoato88:
As discussion [here](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/73#issuecomment-589786484), dashboard-operator had moved to `sigs.k8s.io/cluster-addons` by [PR in cluster-addons](https://github.com/kubernetes-sigs/cluster-addons/pull/44)
Now, we can remove dashboard-operator from `sigs.k8s.io/kubebuilder-declarative-pattern` repository.
The following are affected contents for removing dashboard-operator.

- examples/dashboard-operator
  - just remove `dashboard-operator` folder
- hack/ci/test.sh
  - tests include a test of dashboard-operator 
- hack/smoketest.go
  - sample implementation that use dashboard-operator
- docs/addon/walkthrough
  - contents related to implementation of dashboard-operator

IMHO, It better to create split PRs for each items above.

/assign @atoato88

atoato88:
As of now, related PRs are merged.
I close this issue.

/close

PR #82, opened 2020-04-08
Use guestbook-operator in smoketest

atoato88:
In PR #81, dashboard-operator is removed in this repository.
It is needed that change to use guestbook-operator in `hack/smoketest.go` instead of dashboard-operator.

related to: #80

### DIFF
--- a/examples/guestbook-operator/Dockerfile
+++ b/examples/guestbook-operator/Dockerfile
@@ -13,6 +13,7 @@ RUN go mod download
 COPY main.go main.go
 COPY api/ api/
 COPY controllers/ controllers/
+COPY vendor/ vendor/
 
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
--- a/examples/guestbook-operator/Makefile
+++ b/examples/guestbook-operator/Makefile
@@ -38,6 +38,11 @@ deploy: manifests
 	cd config/manager && kustomize edit set image controller=${IMG}
 	kustomize build config/default | kubectl apply -f -
 
+# Teardown controller in the configured Kubernetes cluster in ~/.kube/config
+# This command does things that can't always re-run so the exit codes are ignored
+teardown: manifests
+	kustomize build config/default | kubectl delete -f - || true
+
 # Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
@@ -50,12 +55,15 @@ fmt:
 vet:
 	go vet ./...
 
+vendor:
+	go mod vendor
+
 # Generate code
 generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile=./hack/boilerplate.go.txt paths="./..."
 
 # Build the docker image
-docker-build: test
+docker-build: vendor test
 	docker build . -t ${IMG}
 
 # Push the docker image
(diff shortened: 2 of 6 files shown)


examples/guestbook-operator/Makefile
```diff
@@ -56,7 +61,9 @@ generate: controller-gen
 
 # Build the docker image
 docker-build: test
+	sed -e "s:^replace://replace:g" go.mod > go.mod.make-docker-build
```
atoato88, line 64: This is a rough approach, I think...
Without this change, `go run smoketest.go` will fail.
Here is log.
```
...
...
Step 5/14 : RUN go mod download
 ---> Running in 3e6c6d104896
go: sigs.k8s.io/kubebuilder-declarative-pattern@v0.0.0-20200317144824-bbf1fb2a4a9a: parsing /go.mod: open /go.mod: no such file or directory
The command '/bin/sh -c go mod download' returned a non-zero code: 1
Makefile:64: recipe for target 'docker-build' failed
make: *** [docker-build] Error 1
```

Any Ideas? or, is this reasonable?

examples/guestbook-operator/Dockerfile
```diff
@@ -3,7 +3,8 @@ FROM golang:1.13 as builder
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
-COPY go.mod go.mod
+#COPY go.mod go.mod
+COPY go.mod.make-docker-build go.mod
```
atoato88, line 7: This relate to my comment at L64 in `examples/guestbook-operator/Makefile`

examples/guestbook-operator/Makefile
johnsonj: Hey @atoato88 ! I'm trying to think of a different solution here, we may not have much of an option though. It's a tricky scenario to build correctly.

johnsonj: The problem is that the build target is dependent on a library that is in the same source tree (both with their own go mod files). The separation is necessary because we don't want examples/.. to be part of the library when a user consumes this.

The dashboard-operator took the following approach:

- Rename the kubebuilder-declarative-pattern in the go.mod to ../../: [src](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/bbf1fb2a4a9a5ab3cf14d3f5f3f06658ca6373f8/examples/dashboard-operator/go.mod#L13) 
- Use vendored dependencies (instead of go modules) in the docker build [src](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/bbf1fb2a4a9a5ab3cf14d3f5f3f06658ca6373f8/examples/dashboard-operator/Dockerfile#L13)
- Update the vendor dependencies as part of docker-build [src](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/bbf1fb2a4a9a5ab3cf14d3f5f3f06658ca6373f8/examples/dashboard-operator/Makefile#L56)

This approach ensures that you are always running the latest code from the main repo when running the example. This is low maintenance, great for testing/development. It would 'feel better' if it didn't require using vendored dependencies. There may be a workaround for that, but I'm not sure.

atoato88: Thank you for helpful comment, I'll change my PR according to comment!! :smile:

atoato88: I've fixed about this.
Your comment solved all concerning.
Thank you!!

examples/guestbook-operator/Makefile
```diff
@@ -38,6 +38,11 @@ deploy: manifests
 	cd config/manager && kustomize edit set image controller=${IMG}
 	kustomize build config/default | kubectl apply -f -
 
+# Teardown controller in the configured Kubernetes cluster in ~/.kube/config
+# This command does things that can't always re-run so the exit codes are ignored
```
johnsonj, line 42: thanks for the comment here!

johnsonj:
Thank you @atoato88 !

/lgtm
/approve

PR #85, opened 2020-04-14
Always close watchers

justinsb:
It's not entirely clear whether it's actually required, but it's
certainly a good defensive practice.


atoato88:
/lgtm

I confirmed how is `Stop()` used by [searching "ResultChan"](https://cs.k8s.io/?q=ResultChan&i=nope&files=&repos=) on https://cs.k8s.io/ .
"Stop()" is usually called in same situations, I Think.

PR #86, opened 2020-04-14
Use log instead of klog

justinsb:
For consistency, we should use log instead of klog in the semver comparison
logic.


atoato88:
/lgtm

Nice catch! :tada:

Issue #56, opened 2019-10-16
kustomize integration


annp1987:
Hi Everyone,

Currently, we can adding labels, changing namespace of manifest before apply them. But I'm not sure there is possible to manipulate manifest from `kustomize build` then apply it? Do you have any advice?

johnsonj:
- [proposal for integration](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/57#issuecomment-559268407)

@annp1987 - Should we assign this bug to you?

annp1987:
@johnsonj ,
Please assign this bug to me. Its my pleasure. :-)

johnsonj:
Thanks @annp1987 ! Let me know if you have any questions/want an early review

annp1987:
Thanks @johnsonj. Could you have a early review for kustomize integration at https://github.com/annp1987/kubebuilder-declarative-pattern/tree/intergrate_with_krusty?

johnsonj:
Hey @annp1987 - sorry for the delay. Your change looks right on track. I don't have any major structural comments. I'm excited to see this one land!

This likely goes without saying, but you will need to keep track of the original filesystem paths and use them when you write to the in memory filesystem (`fSys.WriteFile`)

annp1987:
Thanks @johnsonj. I've just create a PR. Could you take a look at it?

johnsonj:
#66 added the initial support for this feature. We still need tests, documentation, and potentially more feature work for this to be closed out.

johnsonj:
There's additional open issues highlighted in [@atoato88's comment](https://github.com/kubernetes-sigs/cluster-addons/pull/46#issuecomment-597415111)

rajansandeep:
Adding further to [@atoato88's comment](https://github.com/kubernetes-sigs/cluster-addons/pull/46#issuecomment-597415111), if an operator has Kustomize enabled (for example CoreDNS-operator to include the `configMapGenerator`), we can just parse the final yaml built by the `kustomization.yaml`, instead of trying to parse all the files present in the version folder.

johnsonj:
As it is today, the transformation process relies on being able to parse everything as YAML. For example, we have a private registry transform that will go patch all of the image registries. That's something we would like to keep.

[...]

PR #88, opened 2020-04-14
Avoid reusing filesystem, cleanup some error handling

justinsb:
Because we can execute concurrently, we shouldn't reuse the in-memory
filesystem instance; construct a new one for each reconcile.

Also be more rigorous in exiting on errors.


atoato88:
I add link to issue #56 because this related to it.

PR #94, opened 2020-05-11
Add ownerRef into kustomize build

rajansandeep:
The manifest built by Kustomize doesn't include the OwnerRefs. Due to this, the manifest built by Kustomize doesn't reconcile with the operator. 

This PR ensures reconciliation by parsing the manifest created by Kustomize into objects and then injecting the OwnerRef needed for Reconcile.

### DIFF
--- a/pkg/patterns/declarative/reconciler.go
+++ b/pkg/patterns/declarative/reconciler.go
@@ -136,33 +136,12 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 	}
 	var manifestStr string
 
-	if r.IsKustomizeOptionUsed() {
-		// run kustomize to create final manifest
-		opts := krusty.MakeDefaultOptions()
-		k := krusty.MakeKustomizer(fs, opts)
-		m, err := k.Run(objects.Path)
-		if err != nil {
-			log.Error(err, "running kustomize to create final manifest")
-			return reconcile.Result{}, fmt.Errorf("error running kustomize: %v", err)
-		}
-		log.Info("running kustomize to create final manifest")
-		manifestYaml, err := m.AsYaml()
-		if err != nil {
-			log.Error(err, "creating final manifest yaml")
-			return reconcile.Result{}, fmt.Errorf("error converting kustomize output to yaml: %v", err)
-		}
-
-		log.Info("creating final manifest yaml")
-		manifestStr = string(manifestYaml)
-
-	} else {
-		m, err := objects.JSONManifest()
-		if err != nil {
-			log.Error(err, "creating final manifest")
-			return reconcile.Result{}, fmt.Errorf("error creating manifest: %v", err)
-		}
-		manifestStr = m
+	m, err := objects.JSONManifest()
+	if err != nil {
+		log.Error(err, "creating final manifest")
+		return reconcile.Result{}, fmt.Errorf("error creating manifest: %v", err)
 	}
+	manifestStr = m
 
 	extraArgs := []string{"--force"}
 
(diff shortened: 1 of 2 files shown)


atoato88:
/lgtm
Thank you to create PR!

johnsonj:
very well structured change, thank you @rajansandeep !

/approve

pkg/patterns/declarative/reconciler.go
```diff
@@ -136,33 +136,12 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 	}
 	var manifestStr string
 
-	if r.IsKustomizeOptionUsed() {
```
camilamacedo86, line 139: HI @johnsonj and @rajansandeep, 

I'd like to share that this change broke the projects built with Kubebuilder which are using the addons. I am able to check it by running the [the script to gen the testdata](https://github.com/kubernetes-sigs/kubebuilder/blob/master/generate_testdata.sh#L76) which will scaffold a project which will get by default the latest version of this project. 

Now, when we try to run [make generate](https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3-addon/Makefile#L53-L55) with [here](https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v3-addon) it gets stuck. 

c/c @mengqiy @droot @estroz @hasbro17

rajansandeep: > NOTE The same was not faced with the version sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200512162422-ce639cbf6d4c

IIUC, the mentioned version includes changes from this PR.

johnsonj: Good find! @camilamacedo86 can you confirm that it does succeed with `ce639cbf6d4ce402273ad35ecd5f2996758b4f6b` (`sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200512162422-ce639cbf6d4c`)

We've only had docs changes since that commit, so I wouldn't expect anything since then to have broken it.

camilamacedo86: Hi @johnsonj,
 
Yes, I can confirm that it works with `sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200512162422-ce639cbf6d4c`. For more details see the issue: https://github.com/kubernetes-sigs/kubebuilder/issues/1525 and the PR https://github.com/kubernetes-sigs/kubebuilder/pull/1527 which will be a fix for kubebuilder (it is just fixing the version that should be used of this project to avoid this scenario as other breaking changes)

However, if I do not fix the kb project to use the specific version `sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200512162422-ce639cbf6d4c` then, when I run the script to gen the projects it gets stuck when try to use the controller-gen for a project which uses it. The other PR merged here is just a README as you said so, my guess is that it might be related to this one. Sharing to allow we figure out since it might be a problem for any controller project with that tries to use the controller-gen. 

I hope that I could clarify.

camilamacedo86: Hi @johnsonj,
 
Yes, I can confirm that it works with `sigs.k8s.io/kubebuilder-declarative-pattern v0.0.0-20200512162422-ce639cbf6d4c`. 

> We've only had docs changes since that commit, so I wouldn't expect anything since then to have broken it.
Exactly, because of this, I ping here. I did not saw anything after that which could cause the issue. 

So, just to clarifies. The Kubebuilder did not have any change in the period and I just not that I was unable to gen the mock testdata dir with its script. There, IMO the fix would be kept the version fixed. However, it might affect others as well. because of this, I decided to ping. I am trying to figure out as well. 

I hope that I could clarify. Please, feel free to reach me out.

camilamacedo86: Hi @johnsonj and @rajansandeep, 

The issue was with go mod cache. 
After cleaning it as follows worked .. sorted out .. sorry for the false alert :-)
```
$ export GO111MODULE=off
$ go clean --modcache
$ go clean
```

rajansandeep: @camilamacedo86 Thank you for confirming :)

johnsonj: awesome! best kind of fix :). Though this brings up: if we can add any CI to this repo to ensure we don't break kubebuilder I am interested!

camilamacedo86: Trolled by the GO cache.

IMO the solution could be  Kubebuilder used a fixed version: https://github.com/kubernetes-sigs/kubebuilder/pull/1527/files. However, a CI here which uses kb to check the changes also shows a great idea indeed to verify this repo. 👍

Issue #65, opened 2019-12-02
Emit Reconciliation Metrics


johnsonj:
A user should be able to emit metrics on the Prometheus metrics endpoint. Metrics should be written to the [controller-runtime Prometheus Registry](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/metrics/registry.go#L23)

  The library should emit a metric for:

- P1: How many times in the last X seconds has reconciliation happened? How many times has it failed?  
    - Implement as two [CounterVec](https://godoc.org/github.com/prometheus/client_golang/prometheus#CounterVec) metrics. ReconcileCount, ReconcileFailedCount. Use labels that identify the object type (eg CoreDNS) and namespace/name. These can be directly defined in [reconciler.go](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/pkg/patterns/declarative/reconciler.go) 

- P2: How many objects (and what type) are being managed by the operator? 
    - Unclear how to implement this. A GaugeVec (with type/namespace/name labels) may work for this if the reconciler can keep track which types are created/destroyed.

/cc @annp1987

vanou:
I'll take it.
/assign

vanou:
I think this issue got closed.
/close

atoato88:
Thank you to deal this.
I think so too.
/close

PR #98, opened 2020-05-31
Add Reconciliation Metrics Functionality

vanou:
This commit
* adds two metrics
  - Count the number of reconciliation & failure of reconciliation
  - Track the number of objects in manifest

* modifies walkthrough example to mention this metrics functionality

* and deletes
   * reconcilerParams.groupVersionKind option
   * WithGroupVersionKind function

Added two metrics are registered to controller-runtime Prometheus registry

This PR is related to #65 .

### DIFF
--- a/pkg/patterns/declarative/metrics.go
+++ b/pkg/patterns/declarative/metrics.go
@@ -1,0 +1,600 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package declarative
+
+import (
+	"sync"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative/pkg/manifest"
+)
+
+const emptyNamespace = "_Empty_"
+const clusterScoped = "_Cluster_"
+
+const (
+	Declarative = "declarative_reconciler"
+)
+
+const (
+	ReconcileCount   = "reconcile_count"
+	ReconcileFailure = "reconcile_failure_count"
+
+	ManagedObjectsRecord = "managed_objects_record"
+)
+
+var metricsRegisterOnce *sync.Once = &sync.Once{}
+
+var (
+	reconcileCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Subsystem: Declarative,
+		Name:      ReconcileCount,
+		Help:      "How many times reconciliation of K8s objects managed by declarative reconciler occurs",
+	}, []string{"group_version_kind", "namespace", "name"})
+
+	reconcileFailure = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Subsystem: Declarative,
+		Name:      ReconcileFailure,
+		Help:      "How many times reconciliation failure of K8s objects managed by declarative reconciler occurs",
+	}, []string{"group_version_kind", "namespace", "name"})
+
+	managedObjectsRecord = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: Declarative,
+		Name:      ManagedObjectsRecord,
+		Help:      "Track the number of objects in manifest",
+	}, []string{"group_version_kind", "namespace", "name"})
+)
+
+var metricsList = []prometheus.Collector{reconcileCount, reconcileFailure, managedObjectsRecord}
+
+func gvkString(gvk schema.GroupVersionKind) string {
+	if len(gvk.Group) == 0 && gvk.Version == "v1" {
+		return gvk.Version + "/" + gvk.Kind
+	}
+
+	return gvk.Group + "/" + gvk.Version + "/" + gvk.Kind
+}
+
+func namespaced(mgr manager.Manager, gvk schema.GroupVersionKind) (bool, error) {
+	mapping, err := mgr.GetRESTMapper().RESTMapping(gvk.GroupKind(), gvk.Version)
+	if err != nil {
+		return false, err
+	}
+
+	if mapping.Scope.Name() == meta.RESTScopeNameNamespace {
+		return true, nil
+	} else {
+		return false, nil
+	}
+}
+
+type reconcileMetrics struct {
+	groupVersionKind           string
+	reconcileCounterVec        *prometheus.CounterVec
+	reconcileFailureCounterVec *prometheus.CounterVec
+}
+
+func reconcileMetricsFor(gvk schema.GroupVersionKind) reconcileMetrics {
+	return reconcileMetrics{groupVersionKind: gvkString(gvk),
+		reconcileCounterVec: reconcileCount, reconcileFailureCounterVec: reconcileFailure}
+}
+
+func (rm *reconcileMetrics) reconcileWith(req reconcile.Request) {
+	rm.reconcileCounterVec.WithLabelValues(rm.groupVersionKind, req.Namespace, req.Name).Inc()
+}
+
+func (rm *reconcileMetrics) reconcileFailedWith(req reconcile.Request, _ reconcile.Result, err error) {
+	if err != nil {
+		rm.reconcileFailureCounterVec.WithLabelValues(rm.groupVersionKind, req.Namespace, req.Name).Inc()
+	}
+}
+
+type objectRecorder struct {
+	groupVersionKind string
+	gaugeVec         *prometheus.GaugeVec
+}
+
+func objectRecorderFor(gvk schema.GroupVersionKind) objectRecorder {
+	return objectRecorder{groupVersionKind: gvkString(gvk), gaugeVec: managedObjectsRecord}
+}
+
+func (or objectRecorder) Set(namespace, name string, num float64) {
+	if namespace == emptyNamespace || namespace == clusterScoped {
+		or.gaugeVec.WithLabelValues(or.groupVersionKind, "", name).Set(num)
+	} else {
+		or.gaugeVec.WithLabelValues(or.groupVersionKind, namespace, name).Set(num)
+	}
+}
+
+func (or objectRecorder) Unset(namespace, name string) {
+	if namespace == emptyNamespace || namespace == clusterScoped {
+		or.gaugeVec.DeleteLabelValues(or.groupVersionKind, "", name)
+	} else {
+		or.gaugeVec.DeleteLabelValues(or.groupVersionKind, namespace, name)
+	}
+}
+
+var globalObjectTracker *ObjectTracker = NewObjectTracker()
+
+// GetMetricsDuration function returns current metricsDuration
+// of package-scoped internal variable of type *ObjectTracker.
+// It is safe to call this function from multiple go routines
+// with another go routines calling SetMetricsDuration function.
+func GetMetricsDuration() int {
+	return globalObjectTracker.GetMetricsDuration()
+}
+
+// SetMetricsDuration function sets metricsDuration of package
+// scoped internal variable of type *ObjectTracker.
+// It is safe to call this function from multiple go routines
+// with another go routines calling GetMetricsDuration function.
+func SetMetricsDuration(metricsDuration int) {
+	globalObjectTracker.SetMetricsDuration(metricsDuration)
+}
+
+// Type ObjectTracker manages metricsDuration of k8s objects
+// managed by controller(s)
+type ObjectTracker struct {
+	mu              sync.Mutex
+	mgr             manager.Manager
+	metricsDuration int
+	trackedGVK      map[schema.GroupVersionKind]*gvkTracker
+}
+
+// GetMetricsDuration method returns current metricsDuration.
+// It is safe to call this function from multiple go routines
+// with another go routines calling SetMetricsDuration method.
+func (ot *ObjectTracker) GetMetricsDuration() int {
+	ot.mu.Lock()
+	defer ot.mu.Unlock()
+
+	return ot.metricsDuration
+}
+
+// SetMetricsDuration method sets metricsDuration.
+// It is safe to call this function from multiple go routines
+// with another go routines calling GetMetricsDuration method..
+func (ot *ObjectTracker) SetMetricsDuration(metricsDuration int) {
+	ot.mu.Lock()
+	defer ot.mu.Unlock()
+
+	ot.metricsDuration = metricsDuration
+}
+
+func (ot *ObjectTracker) setMetricsDurationInternal(i int) {
+	ot.mu.Lock()
+	defer ot.mu.Unlock()
+
+	if i > 0 && i > ot.metricsDuration {
+		ot.metricsDuration = i
+	}
+}
+
+// Call this before kubectl.Apply
+func (ot *ObjectTracker) addIfNotPresent(objects []*manifest.Object, defaultNamespace string) errors.Aggregate {
+	ot.mu.Lock()
+	defer ot.mu.Unlock()
+	defer ot.deleteMetricsIfNeeded()
+	defer ot.inc()
+
+	var errs []error
+
+	for _, object := range objects {
+		gvk, ns, name := object.GroupVersionKind(),
+			object.UnstructuredObject().GetNamespace(),
+			object.UnstructuredObject().GetName()
+
+		// Check default namespace
+		if defaultNamespace != "" {
+			ns = defaultNamespace
+		}
+
+		namespaced, err := namespaced(ot.mgr, gvk)
+		if err != nil {
+			errs = append(errs, newNoRESTMapperErr(err, gvk))
+			continue
+		}
+
+		if namespaced {
+			if len(ns) == 0 {
+				ns = emptyNamespace
+				errs = append(errs, newEmptyNamespaceErr(gvk, object.UnstructuredObject().GetName()))
+			}
+		} else {
+			ns = clusterScoped
+		}
+
+		if gvkt, ok := ot.trackedGVK[gvk]; ok {
+			if !gvkt.resetIfItHas(ns, name) {
+				gvkt.insert(ns, name)
+			}
+			continue
+		}
+
+		ot.trackedGVK[gvk] = newGVKTracker(ot.mgr, object.UnstructuredObject(), namespaced)
+		ot.trackedGVK[gvk].insert(ns, name)
+
+		// addIfNotPresent is called at Reconcler.reconcileExists,
+		// so Controller & Manager is already running
+		ot.trackedGVK[gvk].start()
+	}
+
+	return errors.NewAggregate(errs)
+}
+
+// Call after ot.mu.Lock()
+func (ot *ObjectTracker) inc() {
+	for _, gvkt := range ot.trackedGVK {
+		gvkt.list.inc()
+	}
+}
+
+// Call after ot.mu.Lock()
+func (ot *ObjectTracker) deleteMetricsIfNeeded() {
+	if ot.metricsDuration > 0 {
+		for _, gvkt := range ot.trackedGVK {
+			gvkt.deleteMetricsIfNeeded(ot.metricsDuration)
+		}
+	}
+}
+
+// NewObjectTracker returns new instance of *ObjectTracker
+func NewObjectTracker() *ObjectTracker {
+	ot := &ObjectTracker{}
+	ot.mgr = nil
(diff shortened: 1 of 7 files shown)


vanou:
Thanks @atoato88 for reviewing!

I updated PR.

pkg/patterns/declarative/metrics.go
```diff
@@ -0,0 +1,272 @@
...
+func (ot objectTracker) addIfNotPresent(object *manifest.Object) {
+	gvk, ns, name := object.GroupVersionKind(),
+		object.UnstructuredObject().GetNamespace(),
+		object.UnstructuredObject().GetName()
+
+	if len(ns) == 0 {
+		ns = DefaultNamespace
+	}
+
+	if gvkt, ok := ot.trackedGVK[gvk]; ok {
```
johnsonj, line 133: I think you'll want to add a lock/mutex around trackedGVK. It is possible to have parallel reconciles if the user enables it in controller-runtime[1]
 
[1] https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller?tab=doc#Options

pkg/patterns/declarative/reconciler.go
```diff
@@ -63,29 +69,44 @@ type DeclarativeObject interface {
...
 	r.mgr = mgr
+	globalObjectTracker.mgr = mgr
 
-	if err := r.applyOptions(opts...); err != nil {
-		return err
+	if err = r.applyOptions(opts...); err != nil {
+		return
+	}
+
+	metrics.Registry.MustRegister(reconcileCount, reconcileFailure, managedObjectsRecord)
```
johnsonj, line 85: Can you make Metrics tracking an optional (default off) feature then update the [example](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/tree/master/docs/addon/walkthrough#using-the-framework-in-the-controller) to turn it on?

This is the pattern we've been following (eg with [WithApplyPrune](https://pkg.go.dev/sigs.k8s.io/kubebuilder-declarative-pattern@v0.0.0-20200522144838-848d48e5b073/pkg/patterns/declarative?tab=doc#WithApplyPrune)). This allows us to keep a stable API/functionality for users while introducing new features.

I expect, not so far in the future, we have a 'Recommended Set of Options' and I would certainly expect this to be apart of that. For now, our recommended set is the ones in the walkthrough.

pkg/patterns/declarative/reconciler.go
```diff
@@ -337,6 +362,43 @@ func (r *Reconciler) validateOptions() error {
...
+				for _, check := range gvks {
+					if gv.Version == check.Version {
+						r.rc = reconcileCounterFor(check)
+						r.rfc = reconcileFailureCounterFor(check)
+						return
+					}
+				}
+			}
+		}
+		// Are there any function to recognize K8s API versioning?
```
johnsonj, line 393: I'm not sure, @justinsb are you aware of anything "upstream" to resolve the api version here?

pkg/patterns/declarative/metrics.go
```diff
@@ -0,0 +1,272 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package declarative
+
+import (
```
johnsonj, line 19: Could you add test coverage for this file?

pkg/patterns/declarative/reconciler.go
```diff
@@ -63,29 +69,44 @@ type DeclarativeObject interface {
 // For mocking
 var kubectl = kubectlcmd.New()
 
-func (r *Reconciler) Init(mgr manager.Manager, prototype DeclarativeObject, opts ...reconcilerOption) error {
+func (r *Reconciler) Init(mgr manager.Manager, prototype DeclarativeObject, opts ...reconcilerOption) (err error) {
```
johnsonj, line 72: super nit: can you drop the named return here and in setReconcileMetrics? it makes it a bit harder to read the function once it's more than few lines.

vanou:
Thanks @johnsonj for review! I'll work on your feedback.

pkg/patterns/declarative/reconciler.go
vanou: OK. I'll drop named return in `*Reconciler.Init` here. However, in `setReconcileMetrics`, we cannot use short variable declaration [here](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/98/files#diff-23c5c12ad7f72be6835d4f216e702363R370) because `gvks` will be redeclared & be in scope of 'if' block.

Should I use function scope variable `err`?

pkg/patterns/declarative/reconciler.go
vanou: Yes.

When
* `Scheme.ObjectKinds` returns multiple `GruopVersionKind` &
* `Scheme.PrioritizedVersionsForGroup` doesn't help determine which version to use,

we need a way to recognize K8s API versioning & choose appropriate (more matured) version.

pkg/patterns/declarative/reconciler.go
```diff
@@ -337,6 +362,43 @@ func (r *Reconciler) validateOptions() error {
...
+			}
+		}
+
+		if vp := scheme.PrioritizedVersionsForGroup(gvks[0].Group); len(vp) != 0 {
+			for _, gv := range vp {
+				for _, check := range gvks {
+					if gv.Version == check.Version {
+						r.rc = reconcileCounterFor(check)
+						r.rfc = reconcileFailureCounterFor(check)
+						return
```
johnsonj, line 388: is return the right  decision here? I'm having a hard time understanding what's going on in this block. Can you add a comment?

pkg/patterns/declarative/reconciler.go
johnsonj: In the remaining cases in setReconcileMetrics you should `return nil`, this makes it clear when reading the code that no error can be returned from this valid path.

vanou: Thanks. I understand.

pkg/patterns/declarative/reconciler.go
vanou: In `setReconcileMetrics`, I would like to know GVK of `Reconciler.prototype` to set metrics' `group_version_kind` label value. However `prototype` argument of `Reconcler.Init` is usually address of empty CRD type and `prototype.GetObjectKind().GroupVersionKind()` doesn't return helpful information.

In this commit, I try to determine GVK from `Scheme`. However there is a chance controller manager's `Scheme` returns multiple version of GVK holding same CRD type. To determine which version to use, I use `PrioritizedVersionsForGroup`. If GVK `scheme.PrioritizedVersionsForGroup` returns matches one of GVKs returned by `scheme.ObjectKinds`, we use that GVK as `group_verison_kind` label value. If not, we use `gvks[0]`.

However, if setting [`reconcilerParams.groupVersionKind`]( https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/pkg/patterns/declarative/options.go#L42) becomes mandatory (`Reconciler.validateOptions` check it), there is no need to work on controller manager's `Scheme` to determine GVK.

johnsonj: Ah! Thank you for the explanation @vanou - You've uncovered an issue here! WithGroupVersionKind[1] states 'This option is required'[1] but there is no validation around that, and no operators are specifying it! (checked this repo and kubernetes-sigs/cluster-addons).

For this code, you should use apiutil.GVKForObject(..)[2] and we should just delete the unused `reconcilerParams.groupVersionKind` / `WithGroupVersionKind` option. 

[1] https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/dfc7666ed8e6583e6a37fe39c08d9471aaa241d0/pkg/patterns/declarative/options.go#L95
[2] https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/ce639cbf6d4ce402273ad35ecd5f2996758b4f6b/pkg/patterns/declarative/reconciler.go#L367

vanou: OK. Thanks!

vanou:
Hi @johnsonj .

I dealt with your feedback. Thank you! In this updated commit, I add one functionality to recognize whether K8s object is namespaced or cluster-scoped.

k8s-ci-robot says *pull-declarative-test* failed. This is because I use [envtest package](https://github.com/kubernetes-sigs/controller-runtime/tree/master/pkg/envtest) in sigs.k8s.io/controller-runtime. This package requires kube-apiserver, etcd & kubectl binary to run local kube-apiserver and against it we can run test.

pkg/patterns/declarative/reconciler.go
```diff
@@ -64,28 +72,69 @@ type DeclarativeObject interface {
...
+			}()
+		} else if err = r.tryToSetCounter(); err == nil {
+			r.rc.reconcileWith(request)
+			defer func() {
+				if err != nil {
+					r.rfc.failureWith(request)
+				}
+			}()
+		} else {
+			log.WithName("declarative_reconciler").Error(err, "failed to get corresponding RESTMapper from API server. try next time")
```
justinsb, line 130: Nit: I'm not sure what "try next time" is implying ... does it mean "will retry" ?

pkg/patterns/declarative/metrics.go
```diff
@@ -0,0 +1,342 @@
...
+		Subsystem: Declarative,
+		Name:      ReconcileFailure,
+		Help:      "How many times reconciliation failure of K8s objects managed by declarative reconciler occurs",
+	}, []string{"group_version_kind", "namespace", "name"})
+
+	managedObjectsRecord = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Subsystem: Declarative,
+		Name:      ManagedObjectsRecord,
+		Help:      "Track the number of objects in manifest",
+	}, []string{"group_version_kind", "namespace", "name"})
```
justinsb, line 69: My understanding is that we should generally avoid unbounded dimensions, because of the limits of prometheus scalability.  But this advice may be a few years out of date ... is it still true?

I don't know if we would classify these as unbounded, because in general they are going to be specified in the manifests.  One possible problem is if we started hashing configmaps (e.g. using kustomize), then the number of names would grow.

I want to see these metrics though - could we / should we start with just the group_version_kind dimension?

pkg/patterns/declarative/metrics.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+
+	if mapping.Scope.Name() == meta.RESTScopeNameNamespace {
+		return true, nil
+	} else {
+		return false, nil
+	}
+}
+
+type reconcileCounter struct {
```
justinsb, line 93: One thing you could here is to have something like `reconcileMetrics` and expose a `ReconcileSuceeded` and `ReconcileFailed` method, and not expose the metric structure externally.  Then e.g. you could have `ReconcileFailed` take an error, if in future you wanted to handle some errors differently (a different metric) you could do that entirely in this metrics layer.

pkg/patterns/declarative/options.go
```diff
@@ -92,15 +92,6 @@ func WithObjectTransform(operations ...ObjectTransform) reconcilerOption {
 	}
 }
 
-// WithGroupVersionKind specifies the GroupVersionKind of the managed custom resource
-// This option is required.
-func WithGroupVersionKind(gvk schema.GroupVersionKind) reconcilerOption {
-	return func(p reconcilerParams) reconcilerParams {
-		p.groupVersionKind = &gvk
```
justinsb, line 99: Nice cleanup - thanks!

pkg/patterns/declarative/reconciler.go
```diff
@@ -391,6 +448,17 @@ func (r *Reconciler) injectOwnerRef(ctx context.Context, instance DeclarativeObj
 	return nil
 }
 
+func (r *Reconciler) tryToSetCounter() error {
+	if mapping, err := r.mgr.GetRESTMapper().RESTMapping(r.gvk.GroupKind(), r.gvk.Version); err == nil {
```
justinsb, line 452: Possible simplification:  rather than have the caller check `setCounter`, you could check in this function for the cached `rc` and `rfc` values, and return them.

Also I think you need a mutex guarding `setCounter` because I think Reconcile can be shared between goroutines.

If you're able to refactor `rc` and `rfc` into the same object, I would just store it in a single field e.g. `reconcileMetrics`.  Then I think we can change this method to be `getReconcileMetrics() (*reconcileMetrics, error)`

pkg/patterns/declarative/reconciler.go
```diff
@@ -64,28 +72,69 @@ type DeclarativeObject interface {
...
+	return nil
 }
 
 // +rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
-func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+func (r *Reconciler) Reconcile(request reconcile.Request) (result reconcile.Result, err error) {
 	ctx := context.TODO()
 	log := log.Log
 
+	if r.CollectMetrics() {
```
justinsb, line 114: Could you replace this with `defer r.collectMetrics(request, result, err)` and isolate the logic into a separate function

pkg/patterns/declarative/reconciler.go
```diff
@@ -143,6 +192,14 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 	}
 	manifestStr = m
 
+	if r.CollectMetrics() {
+		for _, obj := range objects.Items {
+			if err := globalObjectTracker.addIfNotPresent(obj); err != nil {
```
justinsb, line 197: You could move this below the `ns` population logic, and then you can pass in the default namespace into `addIfNotPresent`

pkg/patterns/declarative/metrics.go
```diff
@@ -0,0 +1,342 @@
...
+	if err != nil {
+		return err
+	}
+
+	if namespaced {
+		if len(ns) == 0 {
+			ns = DefaultNamespace
+		}
+	} else {
+		ns = ClusterScoped
```
justinsb, line 183: Does prometheus allow the empty value?  I suspect that would be more intuitive if so.

pkg/patterns/declarative/metrics.go
```diff
@@ -0,0 +1,342 @@
...
+		object.UnstructuredObject().GetName()
+
+	namespaced, err := namespaced(ot.mgr, gvk)
+	if err != nil {
+		return err
+	}
+
+	if namespaced {
+		if len(ns) == 0 {
+			ns = DefaultNamespace
```
justinsb, line 180: You need to take the default namespace as an arg, I think, because it won't always be `default`

pkg/patterns/declarative/metrics.go
```diff
@@ -0,0 +1,342 @@
...
+
+func newObjectTracker() *objectTracker {
+	ot := &objectTracker{}
+	ot.mgr = nil
+	ot.trackedGVK = make(map[schema.GroupVersionKind]*gvkTracker)
+
+	return ot
+}
+
+type gvkTracker struct {
```
justinsb, line 213: This is impressive, but a lot of work.  What do we think this metric will be used for?  Should we track a counter of the number of applies instead (which might be easier, and give similar results, depending on what we want it for)?

justinsb:
Thanks @vanou  - some good stuff.  I had some feedback and some questions, particularly around the cardinality and what we think the per-objects counts will be used for.  I'm wondering if we should just start with something simpler (e.g. a count of apply operations, per gvk)...

vanou:
Thanks @justinsb . I'll work on some of your recommendation.

About what metrics should we have, I think, tracking number of apply per gvk is not useful. This won't tell what objects is deployed by current channel. @johnsonj , do you have any comment for metrics functionality roadmap?

justinsb:
One compromise we could make is to put the applied metrics behind a "flag" - i.e. not always on.  So you (and others) can still have it, but we're not creating the unbounded metrics by default for other operators.  Would that be reasonable?

vanou:
It's reasonable for me. However, in my understanding, `WithReconcileMetrics` function in `options.go` seems enough. Could you tell me what should I do to put metrics behind flag?

pkg/patterns/declarative/reconciler.go
vanou: Yes. This means "will try". I'll modify it to "Reconciler will try to get RESTMapper next time."

pkg/patterns/declarative/metrics.go
vanou: @justinsb 
One approach to this is to introduce aging to pair of `group_version_kind`, `namespace` & `name`. If a triad hasn't been kubectl-applied but reconciliation has happened *n* times, that triad is removed from prometheus metrics. `n` can be set by any number when `WithReconcileMetrics` is called.
Is it reasonable?

pkg/patterns/declarative/reconciler.go
vanou: Thanks. I overlooked that `ns` population logic. I'll do it.

pkg/patterns/declarative/metrics.go
vanou: As your comment on `ns` population logic, you're right. I'll deal with it. Thanks.

pkg/patterns/declarative/metrics.go
vanou: Promtheus allows empty value. However, here, I need something to distinguish namespaced object and cluster-scoped object. 
I'll set metrics namespace label to empty value if namespace is not specified because it isn't reasonable to say empty namespace is `default` namespace as you point out in another comment.

pkg/patterns/declarative/metrics.go
vanou: I'll create new type `reconcileMetrics` in `metrics.go` file and use it. That type will implement similar method you suggest.

pkg/patterns/declarative/metrics.go
```diff
@@ -0,0 +1,342 @@
...
+		return reconcileCounter{groupVersionKind: gvkString(gvk), namespaced: true, counterVec: reconcileCount}
+	} else {
+		return reconcileCounter{groupVersionKind: gvkString(gvk), namespaced: false, counterVec: reconcileCount}
+	}
+}
+
+func (rc reconcileCounter) reconcileWith(request reconcile.Request) {
+	if rc.namespaced {
+		if len(request.Namespace) == 0 {
+			request.Namespace = DefaultNamespace
```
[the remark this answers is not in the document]
vanou: You are right. I'll leave namespace as empty if `reconcile.Request` has empty namespace value. And because of it, I remove code checking whether CRD is namespaced through `RESTMapper`.

pkg/patterns/declarative/reconciler.go
vanou: I'll create type `reconcileMetrics` in `metrics.go` and use it. And I'll stop checking whether CRD is namespaced through `RESTMapper`, because I'll leave namespace as empty value if `reconcile.Request.Namaspace` is empty.

pkg/patterns/declarative/reconciler.go
vanou: OK. I'll do it.

vanou:
Hello @justinsb . I modified PR following your recommendation.
However I added one feature which enables deleting metrics of removed k8s objects after user-specified times reconciliation happened. I think this feature solves unbounded metrics problem.

As I mentioned above, CI test failed because of [envtest package](https://github.com/kubernetes-sigs/controller-runtime/tree/master/pkg/envtest) used in unit test.

Your comment is very helpful. Thank you.

pkg/patterns/declarative/metrics.go
```diff
@@ -0,0 +1,600 @@
...
+}
+
+func newEmptyNamespaceErr(gvk schema.GroupVersionKind, name string) emptyNamespaceErr {
+	return emptyNamespaceErr{gvk: gvk, name: name}
+}
+
+type gvkTracker struct {
+	list     *items
+	src      source.Source
+	evthdler handler.EventHandler
```
justinsb, line 315: Nit: eventHandler and predicate; shorter variable names are harder to read (and probably not _that_ much faster to type any more :-) , particularly with autocomplete)

pkg/patterns/declarative/metrics.go
```diff
@@ -0,0 +1,600 @@
...
+	return b, p
+}
+
+func newItems() *items {
+	it := &items{}
+	it.ids = make(map[string]bol)
+	return it
+}
+
+type bol map[string]*counter
```
justinsb, line 474: Nit: I'm not sure what bol is short for, the full name would be clearer

justinsb:
So as this is optional (thanks for pointing that out), I think it's reasonable to merge this in and we can iterate on it as needed :-)

Some style nits: I don't _think_ we normally abbreviate variable names (obvious exceptions are method receivers, range variables and often the word `obj` for object ! )

/approve
/lgtm

pkg/patterns/declarative/reconciler.go
```diff
@@ -18,6 +18,7 @@ package declarative
 
 import (
 	"context"
+	stdlibErrs "errors"
```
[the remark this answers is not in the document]
vanou: I use `errors` to standard package & use `apierrors` to `k8s.io/apimachinery/pkg/api/errors`. All lowercase as `apiutil` for `sigs.k8s.io/controller-runtime/pkg/client/apiutil`.

pkg/patterns/declarative/metrics.go
vanou: First I use `bol` as abbreviation of `bookOfLife`. However I'll use `record` instead of `bol`.

vanou:
Thanks @justinsb . I modified names of variables.

> So as this is optional (thanks for pointing that out), I think it's reasonable to merge this in and we can iterate on it as needed :-)

*(Just to avoid miscommunication)*
Yes, this feature is optional. However I improved PR to address unbounded metrics problem, so I think it's bit fair to use this feature without concern. (Of course, there may be points we should improve.)

vanou:
Hi @justinsb . I rebased this PR. However, for my fault, this PR contains 4 commits now. Should I do any workaround?

justinsb:
Yeah, would you mind squashing the commits @vanou ?  We don't always have to - it can be nice to divide a big commit into logical steps - but in this case the commits seem mostly to be merging/rebasing, so squashing would be great!

vanou:
> Yeah, would you mind squashing the commits @vanou ? We don't always have to - it can be nice to divide a big commit into logical steps - but in this case the commits seem mostly to be merging/rebasing, so squashing would be great!

Thanks @justinsb . I tried to squash and because of need of resolving merge conflict, it resulted in 2 commits. And I think this PR is ready.

justinsb:
Thanks @vanou :tada: 

/approve
/lgtm

PR #100, opened 2020-06-23
Moves some functions from Coredns operator that could be reused 

SomtochiAma:
…erator


justinsb:
Thanks @SomtochiAma - we might want more package structure in future, but I think we can evolve that (I don't think it's a problem to occasionally break these functions while they are still new)

/approve
/lgtm

PR #101, opened 2020-06-26
Uses new apply function so we don't need kubectl

SomtochiAma:


pkg/patterns/declarative/pkg/kubectlcmd/client.go
```diff
@@ -49,7 +56,8 @@ func (console) Run(c *exec.Cmd) error {
 }
 
 // Apply runs the kubectl apply with the provided manifest argument
-func (c *Client) Apply(ctx context.Context, namespace string, manifest string, validate bool, extraArgs ...string) error {
+func (c *Client) ApplyAlt(ctx context.Context, namespace string, manifest string, validate bool,
+	extraArgs ...string) error {
```
justinsb, line 60: Nit we tend not to split these over lines, or if we do we split all the args, one per line

pkg/patterns/declarative/pkg/kubectlcmd/client.go
```diff
@@ -87,3 +95,41 @@ func (c *Client) Apply(ctx context.Context, namespace string, manifest string, v
...
+
+func(c *Client) Apply(ctx context.Context, namespace string, manifest string, validate bool,
+	extraArgs ...string) error{
+	ioStreams := genericclioptions.IOStreams{
+		In:     os.Stdin,
+		Out:    os.Stdout,
+		ErrOut: os.Stderr,
+	}
+	restClient := genericclioptions.NewConfigFlags(true).WithDeprecatedPasswordFlag()
+	//f := util.NewFactory(restClient)
```
justinsb, line 107: Nit: remove this (I think it's just old code)

pkg/patterns/declarative/pkg/kubectlcmd/client.go
```diff
@@ -87,3 +95,41 @@ func (c *Client) Apply(ctx context.Context, namespace string, manifest string, v
...
+	ioReader := strings.NewReader(manifest)
+
+	b := resource.NewBuilder(restClient)
+	res := b.Unstructured().Stream(ioReader, "cm").Do()
+	//infos = append(infos, namespaceObj)
```
justinsb, line 112: Nit: remove if it's old code

pkg/patterns/declarative/pkg/kubectlcmd/client.go
```diff
@@ -87,3 +95,41 @@ func (c *Client) Apply(ctx context.Context, namespace string, manifest string, v
...
+	infos, err := res.Infos()
+	if err != nil {
+		return err
+	}
+	applyOpts := apply.NewApplyOptions(ioStreams)
+	applyOpts.Namespace = namespace
+	buildApplyOpts(applyOpts, infos, ioStreams)
```
justinsb, line 119: Not sure if splitting this function out is worth it vs just inlining it?

pkg/patterns/declarative/pkg/kubectlcmd/client.go
```diff
@@ -49,7 +56,8 @@ func (console) Run(c *exec.Cmd) error {
 }
 
 // Apply runs the kubectl apply with the provided manifest argument
-func (c *Client) Apply(ctx context.Context, namespace string, manifest string, validate bool, extraArgs ...string) error {
+func (c *Client) ApplyAlt(ctx context.Context, namespace string, manifest string, validate bool,
```
justinsb, line 59: Nit: I think having two types might be clearer here - e.g. one `ExecKubectl` and one called `DirectApply`, that both implement the same interface.  We can discuss in our GSoC sync :-)

justinsb:
Thanks @SomtochiAma - looks great and this is a big win in terms of our kubectl dependency!

/approve
/lgtm

PR #102, opened 2020-07-03
Adds Version Check

SomtochiAma:


pkg/patterns/addon/pkg/status/basic.go
```diff
@@ -29,3 +29,13 @@ func NewBasic(client client.Client) declarative.Status {
 		// no preflight checks
 	}
 }
+
+// NewBasicVersionCheck provides an implementation of declarative.Status that
+// performs version checks for the version of the operator that the manifest requires.
+func NewBasicVersionChecks(client client.Client, version string) declarative.Status {
+	return &declarative.StatusBuilder{
+		ReconciledImpl: NewAggregator(client),
+		VersionCheckImpl:  NewVersionCheck(client, version),
```
justinsb, line 38: Have you run gofmt?  (And I guess we need to add a gofmt check!)

pkg/patterns/addon/pkg/status/version.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"context"
+	"fmt"
+	"github.com/blang/semver"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	addonsv1alpha1 "sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/addon/pkg/apis/v1alpha1"
+
+	//"sigs.k8s.io/controller-runtime/pkg/log"
```
justinsb, line 11: Nit: should we remove?

pkg/patterns/addon/pkg/status/version.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative"
+	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative/pkg/manifest"
+)
+
+// NewVersionCheck provides an implementation of declarative.Reconciled that
```
justinsb, line 17: Oh cool - using a signature inspired by Reconcilers is an interesting pattern to explore!

pkg/patterns/addon/pkg/status/version.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// Look for annotation from any resource with the max version
+	for _, obj := range objs.Items {
+		annotations := obj.UnstructuredObject().GetAnnotations()
+		if versionNeededStr, ok := annotations["addons.k8s.io/operator-version"]; ok {
+			log.WithValues("version", versionNeededStr).Info("Got version, %v")
+
+			versionActual, err := semver.Make(versionNeededStr)
+			if err != nil {
+				log.WithValues("version", versionNeededStr).Info("Unable to convert string to version, skipping this object")
```
justinsb, line 45: Nit: I think you should report this as a `log.Error(err)`, even if we don't return an error.

pkg/patterns/addon/pkg/status/version.go
```diff
@@ -0,0 +1,81 @@
...
+				continue
+			}
+
+			if versionActual.GT(maxVersion) {
+				maxVersion = versionActual
+			}
+		}
+	}
+
+	// TODO(somtochi): Do we want to return an error when the version is invalid or just skip and use the operator?
```
justinsb, line 55: I actually think it would be reasonable to return an error, if the annotation is specified but we can't parse it.  It _could_ be a version schema our operator is too old to understand.  (More likely, it's just a typo).

But ... the user did specify the annotation in their manifest...

I think returning an error is reasonable, it will help users create valid manifests.  A message in the logs is going to get overlooked and users will be surprised when it doesn't work.

pkg/patterns/addon/pkg/status/version.go
```diff
@@ -0,0 +1,81 @@
...
+	operatorVersion, err := semver.Make(p.version)
+	if err != nil {
+		log.WithValues("version", p.version).Info("Unable to convert string to version, skipping check")
```
justinsb, line 58: You could move this to `NewVersionCheck`; I think it's very natural to return an error there if passed a bad version :-)

Related question is: should we refuse to apply manifests that specify invalid minimum versions (i.e. ignore vs error).  I would vote "error" I think, based on my chain of thoughts in this review!

pkg/patterns/declarative/reconciler.go
```diff
@@ -122,6 +122,22 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 	}
 	log.WithValues("objects", fmt.Sprintf("%d", len(objects.Items))).Info("built deployment objects")
 
+	if r.options.status != nil {
+		bool, err := r.options.status.VersionCheck(ctx, instance, objects)
```
justinsb, line 126: Nit: Don't name it `bool` - it's confusing.  e.g. maybe something like `validVersion, err := r.options.status.VersionCheck(...)`

pkg/patterns/declarative/reconciler.go
```diff
@@ -122,6 +122,22 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
+		if  err != nil {
+			if !bool {
+				// r.client isn't exported so can't be updated in version check function
+				r.client.Status().Update(ctx, instance)
```
justinsb, line 130: Nit: I think this can return an error; we need to bubble it up if so.

pkg/patterns/addon/pkg/status/version.go
SomtochiAma: There are also two versions, what happens if we cannot parse the operator's version?

pkg/patterns/addon/pkg/status/version.go
```diff
@@ -0,0 +1,76 @@
...
+	objs *manifest.Objects,
+) (bool, error) {
+	log := log.Log
+	var minOperatorVersion semver.Version
+
+	// Look for annotation from any resource with the max version
+	for _, obj := range objs.Items {
+		annotations := obj.UnstructuredObject().GetAnnotations()
+		if versionNeededStr, ok := annotations["addons.k8s.io/min-operator-version"]; ok {
+			log.WithValues("version", versionNeededStr).Info("Got version requirement addons.k8s.io/operator-version=%v")
```
justinsb, line 43: Nit: this should say `min-operator-version` to match the annotation value

pkg/patterns/declarative/reconciler.go
```diff
@@ -122,6 +130,24 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 	}
 	log.WithValues("objects", fmt.Sprintf("%d", len(objects.Items))).Info("built deployment objects")
 
+	if r.options.status != nil {
+		isValidVersion, err := r.options.status.VersionCheck(ctx, instance, objects)
+		if err != nil {
+			if !isValidVersion {
+				// r.client isn't exported so can't be updated in version check function
+				err := r.client.Status().Update(ctx, instance)
```
justinsb, line 138: So this actually "shadows" the err value we care about.  So on line 142 err will be nil.

Sometimes people use a different name e.g. `updateErr :=...`

But I think a better pattern - when you can use it - is to so `if err := r.client.Status()...; err != nil` as then the scope of the shadowing err should just be the `if` statement (I think!)

justinsb:
One problem with variable shadowing (and a nit), but otherwise lgtm

justinsb:
Thanks @SomtochiAma ... this looks great

/approve
/lgtm
/ok-to-test

PR #103, opened 2020-07-07
Adds function for parsing list

SomtochiAma:


pkg/patterns/declarative/reconciler.go
```diff
@@ -403,3 +411,30 @@ func (r *Reconciler) IsKustomizeOptionUsed() bool {
 func (r *Reconciler) SetSink(sink Sink) {
 	r.options.sink = sink
 }
+
+func parseListKind(infos *manifest.Objects) error{
+		for i, item := range infos.Items{
```
justinsb, line 416: Nit: gofmt needed, I think

pkg/patterns/declarative/reconciler.go
```diff
@@ -403,3 +411,30 @@ func (r *Reconciler) IsKustomizeOptionUsed() bool {
...
+			if item.Kind == "List" {
```
justinsb, line 417: Nit: we should check the group also.  The easy way is to check the APIGroup (which sadly includes the version, but I think that's OK)

pkg/patterns/declarative/reconciler.go
```diff
@@ -403,3 +411,30 @@ func (r *Reconciler) IsKustomizeOptionUsed() bool {
...
+				fmt.Println("haaa!")
```
justinsb, line 418: Nit: please remove debug code :-)

pkg/patterns/declarative/reconciler.go
```diff
@@ -403,3 +411,30 @@ func (r *Reconciler) IsKustomizeOptionUsed() bool {
...
+				obj := item.UnstructuredObject()
+
+				err := obj.EachListItem(func(obj runtime.Object) error{
```
justinsb, line 421: Nice find on `EachListItem` :+1:

pkg/patterns/declarative/reconciler.go
```diff
@@ -403,3 +411,30 @@ func (r *Reconciler) IsKustomizeOptionUsed() bool {
...
+					castItem := obj.(*unstructured.Unstructured)
```
justinsb, line 422: Nit: renaming castItem to `itemUnstructured` may be clearer (similarly, you might rename `obj` to `itemObject`).  (I do like rust's ability to redefine variables with the same name but with a different type, avoids these problems)

pkg/patterns/declarative/reconciler.go
```diff
@@ -403,3 +411,30 @@ func (r *Reconciler) IsKustomizeOptionUsed() bool {
...
+					}
+					infos.Items = append(infos.Items, newObj)
+					return nil
+				})
+
+				if err != nil {
+					return err
+				}
+
+				infos.Items = append(infos.Items[:i], infos.Items[i+1:]...)
```
justinsb, line 435: I think leigh was suggesting that it might be easier to just keep a second list:

```
var out []*manifest.Object

if IsList(obj) {
  out = append(out, extract(obj))
} else {
  out = append(out, obj)
}
return manifest.NewManifest(out)
}
```

I'm not sure what happens to the iteration when you concurrently modify `infos`!

pkg/patterns/declarative/reconciler.go
SomtochiAma: 😅 Will do

justinsb:
Looks great - thanks @SomtochiAma 

/approve
/lgtm
/ok-to-test

PR #104, opened 2020-07-09
Base git repository working

SomtochiAma:


pkg/patterns/addon/pkg/loaders/git.go
```diff
@@ -0,0 +1,130 @@
...
+
+	if r.subDir != "" {
+		name = r.subDir + "/" + name
+	}
+	b, err := r.readURL(name)
+	if err != nil{
+		log.WithValues("path", name).Error(err, "error reading channel")
+		return nil, err
+	}
+	fmt.Println(string(b))
```
justinsb, line 45: Nit: this looks like leftover debug code.  It's OK to log with a very high verbosity (e.g. 4 or 6) instead, but I would probably just delete it - it's most valuable when you're first developing/debugging.

pkg/patterns/addon/pkg/loaders/git.go
```diff
@@ -0,0 +1,130 @@
...
+
+	channel := &Channel{}
+	if err := yaml.Unmarshal(b, channel); err != nil {
+		return nil, fmt.Errorf("error parsing channel %s: %v", name, err)
```
justinsb, line 49: (If you did want to log the bytes, this could be a good place to log it - when something has gone wrong.  If we were debugging this, the first thing we would ask is "what is the contents of channel?")

pkg/patterns/addon/pkg/loaders/git.go
```diff
@@ -0,0 +1,130 @@
...
+	if !allowedManifestId(id) {
+		return nil, fmt.Errorf("invalid manifest id: %q", id)
+	}
+
+	log := log.Log
+	log.WithValues("package", packageName).Info("loading package")
+
+	var filePath string
+	if r.subDir == "" {
+		filePath = fmt.Sprintf("packages/%v/%v/manifest.yaml", packageName, id)
```
justinsb, line 70: If you use `path.Join` you can probably avoid the repetition here: https://play.golang.org/p/-XheAesp5eZ

(Note: path.Join and not filepath.Join, because even on windows we want a url-style path with forward slashes)

pkg/patterns/addon/pkg/loaders/git.go
```diff
@@ -0,0 +1,130 @@
...
+	})
+	if err != nil && err != git.ErrRepositoryAlreadyExists {
+		return nil, err
+	}
+
+	file, err := os.Open(repoDir + "/" + url)
+	if err != nil{
+		return nil, err
+	}
+
```
justinsb, line 103: Missing `defer file.Close()` I think.

I think you can also just do `ioutil.ReadFile` though, which is probably easier

pkg/patterns/addon/pkg/loaders/git.go
```diff
@@ -0,0 +1,130 @@
...
+	repoDir := "/tmp/repo"
+	fmt.Println(r.baseURL)
+	_, err := git.PlainClone(repoDir, false, &git.CloneOptions{
+		URL:               r.baseURL,
+	})
+	if err != nil && err != git.ErrRepositoryAlreadyExists {
+		return nil, err
+	}
+
+	file, err := os.Open(repoDir + "/" + url)
```
justinsb, line 99: Nit: using `filepath.Join` is generally easier & safer when constructing paths (and here, it's a filesystem path, so you want filepath, not path!)

pkg/patterns/addon/pkg/loaders/git.go
```diff
@@ -0,0 +1,130 @@
...
+func parseGitURL(url string) GitRepository{
+	// checks for git:: suffix
+	var subdir string
+	if strings.HasPrefix(url, "git::") {
+		url = strings.TrimPrefix(url, "git::")
+	}
+
+	// checks for subdirectories
+	if strings.Contains(url, ".git//") {
+		newURL := strings.Split(url, ".git//")
```
justinsb, line 121: Nit: if you use `strings.SplitN(url, ".git//", 2)` you will only split on the first `.git//`, and not drop any other segments.

An alternative would be to check that newURL ends up with `len(.) == 2`

(urlTokens or tokens or urlComponents or components might also be a more descriptive name, as it isn't really a URL)

justinsb:
As there's a conflict, a few last nits that it's worth looking at before we merge

justinsb:
This looks great, thanks @SomtochiAma 

/approve
/lgtm
/ok-to-test

justinsb:
Thanks @SomtochiAma  - tests are happy, and I'm happy :-)

/approve
/lgtm

PR #105, opened 2020-07-13
Checks for ignore reconciliation annotations

SomtochiAma:


justinsb:
We discussed in pairing; we agreed we should look for the annotation on the objects in the cluster, by reading them before applying.  (And that same logic is useful in the health check, I believe)

atoato88:
This PR relates to #111 ?
If so, this PR is replaced with it and will close w/o merge?

SomtochiAma:
> This PR relates to #111 ?
> If so, this PR is replaced with it and will close w/o merge?

This PR could still be important. It checks for the ignore annotation on the Custom resource

SomtochiAma:
@justinsb should I close this?

atoato88:
@SomtochiAma 
Sorry for late response.
> It checks for the ignore annotation on the Custom resource

Yes, I think so too, current code of PR is LGTM.

/lgtm

justinsb:
@SomtochiAma and I discussed this; this might be useful in future so we don't just want to close it, but we instead implemented first an annotation on the manifest itself.

/ok-to-test

/hold

PR #108, opened 2020-07-15
Fix goimports and go mod tidy

justinsb:
In preparation for checking this with github actions, fix go mod and goimports so that they will pass the tests.


atoato88:
/lgtm
Thank you for creating this and sorry for delay review..

PR #109, opened 2020-07-16
Modify CI Test to Fetch kubebuilder Release Binary

vanou:
This commit modifies CI test to fetch kubebuilder release binary before test runs.
This enables to run go test using [envtest package from kubernetes-sig/controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/tree/master/pkg/envtest).

### DIFF
--- a/hack/ci/fetch_kubebuilder_release_bin.sh
+++ b/hack/ci/fetch_kubebuilder_release_bin.sh
@@ -1,0 +1,126 @@
+#!/bin/bash
+
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# This shell fetches kubebuilder release binary and puts it under
+# /tmp/kubebuilder_bin/kubebuilder/bin/
+# Then sets KUBEBUILDER_ASSETS to that path to make go test using
+# envtest package in kubernetes-sigs/controller-runtime work properly.
+#
+# This script is largely based on CI script file "chek-everything.sh"
+# of kubernetes-sigs/controller-runtime
+
+# Version of kubebuilder release binary which this script attempts to fetch
+kubebuilder_release_version="2.3.1"
+
+goos=$(go env GOHOSTOS)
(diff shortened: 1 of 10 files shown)


vanou:
Hello @justinsb . I modified CI test code (pull-declarative-test) to fetch kubebuilder release binary to enable go test using envtest package.

Your comment is thankful.

vanou:
I noticed verify-goimports test failed. According to test error message, should I do *go import* against listed go files?

hack/ci/fetch_kubebuilder_release_bin.sh
```diff
@@ -0,0 +1,126 @@
...
+  if [[ !(-x ${dest_dir}/bin/etcd && -x ${dest_dir}/bin/kubectl && -x ${dest_dir}/bin/kube-apiserver) ]]; then
+    echo "There are no expected binaries at ${dset_dir}/bin : etcd, kubectl & kube-apiserver"
+    exit 1
+  fi
+
+  # Set KUBEBUILDER_ASSETS environment variable
+  export KUBEBUILDER_ASSETS="${dest_dir}/bin"
+}
+
+fetch_kb_bin "/tmp"
```
justinsb, line 126: Nit: It might be clearer to source this script and then invoke fetch_kb_bin from the calling script, just so we know that they are going into /tmp.  But just a nit!

hack/ci/fetch_kubebuilder_release_bin.sh
```diff
@@ -0,0 +1,126 @@
...
+  fi
+  
+  archive_file="kubebuilder_${kubebuilder_release_version}_${goos}_${goarch}.tar.gz"
+  
+  mkdir -p ${dest_dir}
+
+  # Fetch kubebuilder release binary
+  curl -fsL "${fetch_url_base}/${archive_file}" -o "/tmp/${archive_file}"
+  if [[ ${?} != 0 ]]; then
+    echo "Failed to curl kubebuilder release binary"
```
justinsb, line 92: Nit: I do wonder whether it would be easier to just log "is kubebuilder supported on ${goos}_${goarch}" here and skip the per-platform tests !

hack/ci/fetch_kubebuilder_release_bin.sh
```diff
@@ -0,0 +1,126 @@
...
+    echo "Maybe, kubebuilder release binary got broken & doesn't match hash value in ${checksum_file}"
+    exit 1
+  fi
+  )
+
+  # Unpack
+  tar -C ${dest_dir} -xzf "/tmp/${archive_file}" --strip-components=1
+  
+  # Check expected binaries exist
+  if [[ !(-x ${dest_dir}/bin/etcd && -x ${dest_dir}/bin/kubectl && -x ${dest_dir}/bin/kube-apiserver) ]]; then
```
justinsb, line 117: I do wonder if we're going to get away with just apiserver, or whether we're going to need kind (or similar).

justinsb:
Thanks @vanou - looks great :-)

/approve
/lgtm

PR #110, opened 2020-07-20
Uses kstatus to compute status of CR

SomtochiAma:


pkg/patterns/declarative/reconciler.go
```diff
@@ -164,6 +166,26 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 		return reconcile.Result{}, fmt.Errorf("error applying manifest: %v", err)
 	}
 
+	statusMap := make(map[string]bool)
+	for _, object := range objects.Items {
+		res, err := status.Compute(object.UnstructuredObject())
+		if err != nil {
+			// Do we return errors or just skipp
```
justinsb, line 173: Probably log and skip this object; maybe post a status of "unknown" or "error"

pkg/patterns/declarative/reconciler.go
```diff
@@ -164,6 +166,26 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
+		}
+		if res != nil {
+			log.Info("%v %v has status %v, message: %v\n", object.Kind, object.Name, res.Status, res.Message)
+			statusMap[res.Status.String()] = true
```
justinsb, line 177: You could make statusMap `map[status.Status]bool` and avoid the `String()` conversion.  That said, I bet they're strings anyway...

pkg/patterns/declarative/reconciler.go
```diff
@@ -173,6 +183,43 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
+
+		unstruct, err := getObjectFromCluster(object, r.config, r.dynamicClient)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		res, err := status.Compute(unstruct)
+		if err != nil {
+			log.WithValues("unstuctured", unstruct).Error(err, "Unable to get status of unstructured")
+			statusMap[status.NotFoundStatus] = true
```
justinsb, line 197: Is NotFoundStatus the best match?  I'm wondering if we have an "internal error" status

pkg/patterns/declarative/reconciler.go
```diff
@@ -173,6 +183,43 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
+		}
+		if res != nil {
+			log.WithValues("kind", object.Kind).WithValues("name", object.Name).WithValues("status", res.Status).WithValues(
```
justinsb, line 200: This might get a little verbose - we can try it and see!

pkg/patterns/declarative/reconciler.go
```diff
@@ -446,3 +493,41 @@ func parseListKind(infos *manifest.Objects) (*manifest.Objects, error) {
 
 	return &ret, nil
 }
+
+
+func aggregateStatus(m map[status.Status]bool) string {
+	inProgress := m["InProgress"]
```
justinsb, line 499: Nit: can we use the status.Status constants here, instead of the strings?

pkg/patterns/declarative/reconciler.go
```diff
@@ -446,3 +493,41 @@ func parseListKind(infos *manifest.Objects) (*manifest.Objects, error) {
...
+	}
+
+	if failed {
+		return "Failed"
+	}
+	return "Current"
+}
+
+func getObjectFromCluster(obj *manifest.Object, config *rest.Config, dynamicClient dynamic.Interface) (*unstructured.
+	Unstructured, error) {
```
justinsb, line 515: Personally I'd prefer a longer line vs breaking in the middle of a type

pkg/patterns/declarative/reconciler.go
```diff
@@ -446,3 +493,41 @@ func parseListKind(infos *manifest.Objects) (*manifest.Objects, error) {
...
+	getOptions := metav1.GetOptions{}
+	gvk := obj.GroupVersionKind()
+	restMapper, err := apiutil.NewDiscoveryRESTMapper(config)
```
justinsb, line 518: Nit: we might want to cache this in future :-)

pkg/patterns/declarative/reconciler.go
```diff
@@ -446,3 +493,41 @@ func parseListKind(infos *manifest.Objects) (*manifest.Objects, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to create RESTMapper from config: %v", err)
+	}
+	mapping, err := restMapper.RESTMapping(obj.GroupKind(), gvk.Version)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get mapping for resource: %v", err)
+	}
+	ns := obj.UnstructuredObject().GetNamespace()
+	unstruct, err := dynamicClient.Resource(mapping.Resource).Namespace(ns).Get(context.Background(),
```
justinsb, line 527: Nit: always better to pass in the context if you control the method signature... even if the caller just specifies `context.TODO()`, it just avoids having to refactor in future.

pkg/patterns/declarative/reconciler.go
```diff
@@ -446,3 +493,41 @@ func parseListKind(infos *manifest.Objects) (*manifest.Objects, error) {
...
+		obj.Name, getOptions)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get mapping for resource: %v", err)
```
justinsb, line 530: Nit: `unable to get resource: %v`

justinsb:
A few nits, looks great overall!

pkg/patterns/declarative/reconciler.go
SomtochiAma: There is `unknown` status

pkg/patterns/declarative/reconciler.go
SomtochiAma: Should we just log the `kind` and `name` without the `status`

pkg/patterns/declarative/reconciler.go
SomtochiAma: I am not completely sure I understand

pkg/patterns/declarative/reconciler.go
```diff
@@ -173,6 +190,43 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
+	for _, object := range objects.Items {
+
+		unstruct, err := getObjectFromCluster(object, r)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		res, err := status.Compute(unstruct)
+		if err != nil {
+			log.WithValues("unstuctured", unstruct).Error(err, "Unable to get status of unstructured")
```
justinsb, line 203: nit:typo

pkg/patterns/declarative/reconciler.go
justinsb: Unknown is probably better - we don't want other systems to draw incorrect conclusions and NotFound feels very transient.  Another option would be to post no status at all for the whole CR, but I feel like unknown is better.

pkg/patterns/declarative/reconciler.go
justinsb: It's just that you're going to be logging for every "child" object each time you go round the reconcile loop.  I think it's fine to leave in for now, because it will help us understand what's going on, and it's also going to be very obvious; we'll just likely want to reduce it to V(2) once we're all happy that it works as we expect!

pkg/patterns/declarative/reconciler.go
```diff
@@ -173,6 +190,43 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
+			statusMap[status.NotFoundStatus] = true
+		}
+		if res != nil {
+			log.WithValues("kind", object.Kind).WithValues("name", object.Name).WithValues("status", res.Status).WithValues(
+				"message", res.Message).Info("Got status of resource:")
+			statusMap[res.Status] = true
+		}
+	}
+
+	addonObject, ok := instance.(addonsv1alpha1.CommonObject)
```
justinsb, line 213: I've realized we should probably have some helper functions for this, so that we can deal with unstructured also!  We might also be able then to eliminate the need for the methods and just use reflection instead, though I'm not sure if that actually helps us that much.

justinsb:
Looking good, I think we can likely merge once conflicts & tests are resolved.

/ok-to-test

justinsb:
Thanks @SomtochiAma  - will be great to try this out with everything else and see how it behaves!

/approve
/lgtm

justinsb:
Thanks again!

/approve
/lgtm

PR #111, opened 2020-07-20
Checks for annotation for objects in-cluster

SomtochiAma:


pkg/patterns/declarative/reconciler.go
```diff
@@ -134,6 +136,40 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+
+	// dynamic config
+	dynamicClientset, err := dynamic.NewForConfig(r.config)
```
justinsb, line 141: We may want to construct this earlier and cache it in the Reconciler object; though it doesn't matter as much for the dynamic client which doesn't have as much state.

pkg/patterns/declarative/reconciler.go
```diff
@@ -134,6 +136,40 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
+		log.Error(err,"Unable to create dynamic client")
+		return reconcile.Result{}, err
+	}
+
+	newItems := []*manifest.Object{}
+	for _, obj := range objects.Items {
+
+		// Uses unsafe method?? Is it safe?
+		getOptions := metav1.GetOptions{}
+		gvk, _ := meta.UnsafeGuessKindToResource(obj.GroupVersionKind())
```
justinsb, line 152: Nit: rename gvk to `resource` or similar

pkg/patterns/declarative/reconciler.go
```diff
@@ -134,6 +136,40 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
+	dynamicClientset, err := dynamic.NewForConfig(r.config)
+	if err != nil {
+		log.Error(err,"Unable to create dynamic client")
+		return reconcile.Result{}, err
+	}
+
+	newItems := []*manifest.Object{}
+	for _, obj := range objects.Items {
+
+		// Uses unsafe method?? Is it safe?
```
justinsb, line 150: I think you can use RESTMapping, like here https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/pkg/patterns/declarative/pkg/watch/dynamic.go#L64

pkg/patterns/declarative/reconciler.go
```diff
@@ -134,6 +136,40 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
+		getOptions := metav1.GetOptions{}
+		gvk, _ := meta.UnsafeGuessKindToResource(obj.GroupVersionKind())
+		ns := obj.UnstructuredObject().GetNamespace()
+		unstruct, err := dynamicClientset.Resource(gvk).Namespace(ns).Get(context.Background(),
+			obj.Name, getOptions)
+		if err != nil{
+			log.WithValues("name", obj.Name).Error(err, "Unable to get resource")
```
justinsb, line 157: We probably also special-case `apierrors.IsNotFound(err)`; that's going to be the error for every object for a new CR and it's not work logging for IMO.

pkg/patterns/declarative/reconciler.go
```diff
@@ -134,6 +136,40 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
+		}
+		if unstruct != nil {
+			annotations := unstruct.GetAnnotations()
+			if ignoreAnnotation, ok := annotations["addons.operators.ignore"]; ok {
```
justinsb, line 161: Let's discuss a name for this annotation in pairing today!

pkg/patterns/declarative/reconciler.go
```diff
@@ -134,6 +136,40 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
+				if ignoreAnnotation == "true" {
```
justinsb, line 162: We can just use an empty value for the annotation i.e. just check `if ok`.  yaml makes values like "true" annoyingly difficult to set...

pkg/patterns/declarative/reconciler.go
```diff
@@ -134,6 +136,40 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
 	}
+
+	// dynamic config
+	dynamicClientset, err := dynamic.NewForConfig(r.config)
+	if err != nil {
+		log.Error(err,"Unable to create dynamic client")
+		return reconcile.Result{}, err
+	}
+
+	newItems := []*manifest.Object{}
```
justinsb, line 147: Just FYI, you can just do `var newItems []*manifest.Object` - the nil slice is treated like an empty array by append.  (Though if you look at my older code, I did this too!)

pkg/patterns/declarative/reconciler.go
```diff
@@ -143,6 +150,38 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+
+	var newItems []*manifest.Object
+	for _, obj := range objects.Items {
+
+		getOptions := metav1.GetOptions{}
+		gvk := obj.GroupVersionKind()
+		restMapper, err := apiutil.NewDiscoveryRESTMapper(r.config)
```
justinsb, line 159: As with the other PR, probably caching this (like you do dynamicClient) is a win

justinsb:
/ok-to-test

/approve
/lgtm

One potential improvement: caching the RESTMapper, but LGTM!

justinsb:
/test pull-declarative-test

(The tests aren't failing for me locally, so not sure what's going on)

The goimports failure does look legitimate, though

justinsb:
Ah: It should be apierrors, when you rebase on latest origin/master the tests do fail.

justinsb:
Looks good - I think all your PRs are overlapping with each other, and this one is now failing on the goimports check!

/approve
/lgtm

(though you'll need to fix the test anyway)

PR #112, opened 2020-07-22
Unstructured support in different functions

SomtochiAma:


pkg/patterns/addon/application.go
```diff
@@ -42,8 +43,22 @@ const (
...
+	var addonObject addonsv1alpha1.CommonObject
+
+	if unstruct, ok := instance.(*unstructured.Unstructured); ok {
+		v, _, err := unstructured.NestedString(unstruct.Object, "spec", "version")
+		if err != nil {
+			return fmt.Errorf("unable to get version from unstuctured: %v", err)
+		}
+		version = v
+
+		healthy, _, err = unstructured.NestedBool(unstruct.Object, "spec", "healthy")
```
justinsb, line 57: I think this should be "status" "healthy".  We probably also should check the error (as you do with version)

pkg/patterns/addon/patch.go
```diff
@@ -35,25 +35,38 @@ import (
 func ApplyPatches(ctx context.Context, object declarative.DeclarativeObject, objects *manifest.Objects) error {
 	log := log.Log
 
-	p, ok := object.(addonsv1alpha1.Patchable)
-	if !ok {
-		return fmt.Errorf("provided object (%T) does not implement Patchable type", object)
-	}
-
+	var p addonsv1alpha1.Patchable
```
justinsb, line 38: Nit: Do you need to declare this here?  Looks like you could declare it in on line 54 with `else if p, ok := object....` ?

pkg/patterns/addon/patch.go
```diff
@@ -35,25 +35,38 @@ import (
...
-	for _, p := range p.PatchSpec().Patches {
-		// Object is nil, Raw  is populated (with json, even when input was yaml)
-		r := bytes.NewReader(p.Raw)
-		decoder := yaml.NewYAMLOrJSONDecoder(r, 1024)
-		patch := &unstructured.Unstructured{}
+	unstruct, ok := object.(*unstructured.Unstructured)
+	if ok {
+		patch, _, err := unstructured.NestedSlice(unstruct.Object, "spec", "patches")
+		if err != nil {
+			return fmt.Errorf("unable to get patches from unstruct: %v", err)
```
justinsb, line 45: Nit: unstructured, not unstruct, as we expect this will eventually bubble-up to the user.

pkg/patterns/addon/pkg/status/aggregate.go
```diff
@@ -46,50 +47,90 @@ type aggregator struct {
 func (a *aggregator) Reconciled(ctx context.Context, src declarative.DeclarativeObject, objs *manifest.Objects) error {
 	log := log.Log
 
-	instance, ok := src.(addonv1alpha1.CommonObject)
-	if !ok {
+	unstruct, ok := src.(*unstructured.Unstructured)
```
justinsb, line 50: It might be clearer if you constructed the `healthy` state first, and then applied it to the object.  i.e. move the type-checking on src to around line 89

pkg/patterns/addon/pkg/status/aggregate.go
```diff
@@ -46,50 +47,90 @@ type aggregator struct {
...
+		unstructStatus["Errors"] = statusErrors
+		s, _, err := unstructured.NestedMap(unstruct.Object, "status")
 		if err != nil {
-			log.Error(err, "updating status")
-			return err
+			log.Error(err, "getting status")
+			return fmt.Errorf("unable to get status from unstructured: %v", err)
+		}
+		if !reflect.DeepEqual(status, s) {
+			err = unstructured.SetNestedField(unstruct.Object, statusHealthy, "status", "healthy")
```
justinsb, line 113: Missed error-check here

pkg/patterns/addon/pkg/status/aggregate.go
```diff
@@ -46,50 +47,90 @@ type aggregator struct {
...
+			err = unstructured.SetNestedStringSlice(unstruct.Object, statusErrors, "status", "errors")
+			if err != nil {
+				log.Error(err, "updating status")
+				return fmt.Errorf("unable to set status in unstructured", err)
+			}
+
+			log.WithValues("name", unstruct.GetName()).WithValues("status", status).Info("updating status")
+
+			err = a.client.Update(ctx, unstruct)
```
justinsb, line 122: If you make this `a.client.Update(ctx, src)`, can you eliminate the duplication between the two if branches?

justinsb:
/ok-to-test

A few nits, but it's exciting to see this.

pkg/patterns/addon/pkg/status/aggregate.go
SomtochiAma: I don't know if this would work as the status wasn't changed on `src`. Unlike the `SetNestedField` in `unstruct` and `SetCommonStatus` in `instance` that changes the status

pkg/patterns/addon/application.go
```diff
@@ -42,8 +43,25 @@ const (
 
 // TransformApplicationFromStatus modifies the Application in the deployment based off the CommonStatus
 func TransformApplicationFromStatus(ctx context.Context, instance declarative.DeclarativeObject, objects *manifest.Objects) error {
-	addonObject, ok := instance.(addonsv1alpha1.CommonObject)
-	if !ok {
+	var version string
+	var healthy bool
+	var addonObject addonsv1alpha1.CommonObject
+
+	if unstruct, ok := instance.(*unstructured.Unstructured); ok {
```
justinsb, line 50: Looking at the other PRs, we might want to create helper functions here.  e.g. `GetVersion` and `GetHealthy` (or maybe `GetCommonSpec` and `GetCommonStatus`)

pkg/patterns/addon/pkg/status/aggregate.go
justinsb: Do we know where we don't change `src`?  `src` and `unstruct` are just a pointer-cast, so they should be the same object.

(Also, if we're changing status we might have to use `client.Status().Update(...)`

justinsb:
One query about unstruct != src (and one suggestion for future PRs to refactor out some helper methods) but this lgtm.

Looks like we need a goimports fix though

/approve
/lgtm

pkg/patterns/addon/application.go
SomtochiAma: I could definitely work on this. I think it is a good idea

SomtochiAma:
> Do we know where we don't change src? src and unstruct are just a pointer-cast, so they should be the same object.

I will check on that. When we update src/unstruct, do we update the main instance?

pkg/patterns/addon/pkg/status/aggregate.go
SomtochiAma: It seems every time we run the update function, `src` has changed.  I will change to `client.Status().Update(...)` in another PR

PR #113, opened 2020-07-22
WIP: Bring up kind in CloudBuild, to enable smoketests

justinsb:


justinsb:
(I am currently thinking that Github Actions gives a nice experience here, and we may not need CloudBuild support also)

PR #116, opened 2020-08-11
Adds helper functions

SomtochiAma:


pkg/patterns/addon/pkg/loaders/fs.go
```diff
@@ -66,29 +65,12 @@ func (c *ManifestLoader) ResolveManifest(ctx context.Context, object runtime.Obj
...
-
-		spec := addonObject.CommonSpec()
-		version = spec.Version
-		channelName = spec.Channel
-	} else {
-		return nil, fmt.Errorf("object %T was not an addonsv1alpha1.CommonObject", object)
+	version, err := utils.GetCommonVersion(object)
+	componentName, err = utils.GetCommonName(object)
+	channelName, err = utils.GetCommonChannel(object)
+	// All the errors above are checking the same thing? Is there a better way?
```
justinsb, line 71: There is a pattern that is sometimes used in parsing https://blog.golang.org/errors-are-values

However in this case because the methods are different it's better just to repeat the `if err != nil` block 3 times.

It might be we combine GetCommonName / GetCommonVersion / GetCommonChannel into GetCommonSpec in future, in which case we'd only have one error check.

justinsb:
One note on the error checking, but otherwise lgtm.  Looks like a small conflict on one of the files, but in advance of fixing that and the error checking...

/approve
/lgtm

pkg/patterns/addon/pkg/loaders/fs.go
SomtochiAma: Great article! I always wondered about the frequency of `if err != nil `.

justinsb:
Thanks @SomtochiAma 

/ok-to-test
/approve
/lgtm

PR #118, opened 2020-08-19
Refactors code to set and get status from object

SomtochiAma:


pkg/patterns/addon/pkg/apis/v1alpha1/common_types.go
```diff
@@ -47,9 +47,11 @@ type CommonSpec struct {
 
 // CommonSpec is a set of status attributes that must be exposed on all addons.
 type CommonStatus struct {
-	Healthy bool     `json:"healthy"`
-	Errors  []string `json:"errors,omitempty"`
-	Phase   string   `json:"phase,omitempty"`
+	VersionCheckErrors []string
+	StatusErrors       []string
```
justinsb, line 51: Is this intended?  Should we add these two fields in a separate PR?

pkg/patterns/addon/pkg/utils/helpers.go
```diff
@@ -9,48 +9,71 @@ import (
 	addonsv1alpha1 "sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/addon/pkg/apis/v1alpha1"
 )
 
-func GetCommonVersion(instance runtime.Object) (string, error) {
+func genError(v runtime.Object) error {
+	return fmt.Errorf("instance %T is not addonsv1alpha1.CommonObject or unstructured", v)
```
justinsb, line 13: I probably would have just inlined this function each time, but I guess it's nice that it is consistent! :+1:

pkg/patterns/declarative/reconciler.go
```diff
@@ -189,6 +191,25 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 				log.Error(err, "failed to reconcile status")
 			}
 		}
+
+		status, err := utils.GetCommonStatus(instance)
+		if err != nil {
+			log.Error(err, "failed to get status")
+		}
+		compiledStatus := utils.CompileErrors(status)
+		fmt.Println("compiling errors")
```
justinsb, line 200: Should be `log`, rather than using fmt which will go to stdout.

pkg/patterns/addon/pkg/utils/helpers.go
```diff
@@ -61,6 +84,15 @@ func GetCommonName(instance runtime.Object) (string, error) {
...
 	default:
-		return "", fmt.Errorf("instance %T is not addonsv1alpha1.CommonObject or unstructured", v)
+		return "", genError(v)
 	}
 }
+
+func CompileErrors(status addonsv1alpha1.CommonStatus) addonsv1alpha1.CommonStatus {
+	fmt.Println(status.StatusErrors)
+	status.Errors = []string{}
+	status.Errors = append(status.Errors, status.VersionCheckErrors...)
```
justinsb, line 94: I'm a little uneasy about introducing this in this refactor PR

justinsb:
Is it possible to split the version errors into a different PR and keep this just as a refactor?  (I can show you how I would do it in git if you'd like...)

There's also some ongoing discussions about more structured error handling: https://github.com/kubernetes/community/pull/4521 so I think we should be careful when introducing new status!

pkg/patterns/addon/pkg/apis/v1alpha1/common_types.go
```diff
@@ -47,9 +47,11 @@ type CommonSpec struct {
 
 // CommonSpec is a set of status attributes that must be exposed on all addons.
```
atoato88, line 48: nit: replace `CommonSpec` with `CommonStatus`

pkg/patterns/addon/pkg/utils/helpers.go
SomtochiAma: Thanks. I could change it if that would be more concise.

pkg/patterns/declarative/reconciler.go
SomtochiAma: Sure. I will changing it

SomtochiAma:
> Is it possible to split the version errors into a different PR and keep this just as a refactor? (I can show you how I would do it in git if you'd like...)
> 
> There's also some ongoing discussions about more structured error handling: [kubernetes/community#4521](https://github.com/kubernetes/community/pull/4521) so I think we should be careful when introducing new status!

Sure. I will separate it into different PR and then we can discuss the change there

pkg/patterns/addon/pkg/status/kstatus.go
```diff
@@ -47,41 +46,21 @@ func (k *kstatusAggregator) Reconciled(ctx context.Context, src declarative.Decl
...
-	} else {
-		return fmt.Errorf("instance %T was not an addonsv1alpha1.CommonObject or unstructured."+
-			"Unstructured",
-			src)
-	}
-
-	if changed == true {
 		log.WithValues("name", src.GetName()).WithValues("phase", aggregatedPhase).Info("updating status")
-		err := k.client.Status().Update(ctx, src)
+		err = k.client.Status().Update(ctx, src)
```
justinsb, line 63: Nit - if you're not going to need error out side the block, it's probably better to do either `err := ...` or `if err := foo(); err != nil`

justinsb:
Thanks @SomtochiAma - looks great :-)

/approve
/lgtm

PR #119, opened 2020-08-20
Allows use of private repositories

SomtochiAma:


pkg/patterns/addon/pkg/loaders/git.go
```diff
@@ -122,3 +142,72 @@ func parseGitURL(url string) GitRepository {
...
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getAuthMethod() (transport.AuthMethod, error) {
+	sshFile := fmt.Sprintf("%s/.ssh/id_rsa", os.Getenv("HOME"))
```
justinsb, line 188: In the future, we could also support setting this directly in an env var (e.g. `GIT_SSH_KEY` or something like that).  But this is good

justinsb:
This looks great - thanks @SomtochiAma 

/approve
/lgtm

pkg/patterns/addon/pkg/loaders/git.go
SomtochiAma: That sounds great! I will do so in a follow-up PR.

PR #127, opened 2020-10-27
deps: bump to controller-runtime v0.7.0-alpha.5

estroz:
* go.mod,go.sum: bump k8s deps to v0.19.2, controller-runtime to v0.7.0-alpha.5
* pkg: use client.Object instead of runtime.Object and add context.Context to relevant methods

Once v0.7.0 is released we'll need to bump again, but it should be much simpler then.


go.mod
```diff
@@ -4,22 +4,23 @@ go 1.13
...
+	k8s.io/api v0.19.2
+	k8s.io/apimachinery v0.19.2
+	k8s.io/cli-runtime v0.19.2
+	k8s.io/client-go v0.19.2
 	k8s.io/klog v1.0.0
-	k8s.io/kubectl v0.18.4
+	k8s.io/kubectl v0.19.2
 	sigs.k8s.io/cli-utils v0.16.0
-	sigs.k8s.io/controller-runtime v0.6.0
+	sigs.k8s.io/controller-runtime v0.7.0-alpha.4
```
camilamacedo86, line 23: @estroz IMO we need wait for the stable release to be able to bump the dep.

estroz: Is there a good reason to wait? I don't expect any changes to go in aside from bugfixes at this point, although it is technically an alpha release. Unless there is a plan to release this project soon, bumping to alpha now then bumping to stable before a release makes sense to me since the alpha bump will only ever be on master. This also helps test out the controller-runtime v0.7.0 release.

camilamacedo86 (approved):
/lgtm

go.mod
camilamacedo86: We discussed that. I think it is fine we move forward. 
Really thank you for your collab.

estroz:
ping @justinsb

estroz:
@pwittrock can you take a look at this

pwittrock:
@estroz ping me on slack for reviews -- I get too many GH notifications for this to break through the noise

PR #129, opened 2020-11-09
WIP: Support apply-ing to a remote cluster

justinsb:
This allows for management cluster scenarios, where the CRD is
declared in a different cluster from the target.


atoato88:
Oh sorry, as of now WIP status..
Current code is LGMT :smile: 
If remove WIP tag, I'll review again.
/lgtm cancel

christopherhein:
This is great, I'm very excited to try using this in CAPN to apply standard configs to nested clusters since we'll have kubeconfigs in secrets already. Implementation-wise, I'm slightly confused as to how I would target specific clusters though.

PR #134, opened 2020-11-17
github actions tests: replace set-env with GITHUB_ENV

justinsb:
set-env is deprecated for security reasons, so move to the new GITHUB_ENV approach.


rajansandeep (approved):
/lgtm

PR #135, opened 2020-12-07
Don't run transform twice on kustomize inputs

justinsb:
Running object transforms on the raw kustomize files isn't consistent
with how they are normally used, but it also is very difficult to
pre-transform kustomize inputs - for example if we remove an object,
it will likely still be referenced from kustomization.yaml


rajansandeep (approved):
/lgtm

PR #136, opened 2020-12-07
fakeClient should always return NotFound error

justinsb:
Although it's a fake client, we want it to behave like the real one!

### DIFF
--- a/pkg/test/mocks/client_test.go
+++ b/pkg/test/mocks/client_test.go
@@ -1,0 +1,79 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


pkg/test/mocks/client_test.go
```diff
@@ -0,0 +1,79 @@
...
+	client := NewClient(scheme)
+
+	node := v1.Node{}
+	err := client.Get(ctx, types.NamespacedName{Name: "foo"}, &node)
+	if !apierrors.IsNotFound(err) {
+		t.Fatalf("expected IsNotFound error from non-existent get, got %v", err)
+	}
+}
+
+func TestCreateAndGetGet(t *testing.T) {
```
atoato88, line 46: Is `Get` duplicated?

```suggestion
func TestCreateAndGet(t *testing.T) {
```

atoato88 (commented):
Almost LGTM.
Just one comment. it may be nit. :thinking:

pkg/test/mocks/client_test.go
justinsb: Whoops - thanks!

rajansandeep (approved):
/lgtm

Issue #137, opened 2020-12-09
Cannot apply CRD and a CR simultaneously


willbeason:
How to recreate:

1. Create a manifest which includes both a CRD and a CR for that CRD
2. Install both of those via a Kubebuilder controller simultaneously

This happens because of a recent change to how kubectl works in 1.18. Specifically, `k8s.io/cli-runtime/pkg/resource/result.go` now returns an error if the user tries to apply a type that does not already exist on the cluster, _even if the type's CRD is in the current manifest being traversed_.

This was resolved within the kubectl CLI here: https://github.com/kubernetes/kubernetes/pull/89864

However, that resolution is at a higher level than `kubebuilder-declarative-pattern` uses. [direct.go](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/pkg/patterns/declarative/pkg/applier/direct.go) uses `result.go` directly, rather than using the wrapper that the kubectl CLI uses. So the below block:
```
	infos, err := res.Infos()
	if err != nil {
		return err
	}
```
will return errors if `res` contains a CRD and it's CR, if the CRD is not already installed on the cluster. Instead, execution should continue so long as infos is non-empty. It should be still fine to log/return the error, but the call to `applyOpts.Run()` shouldn't be blocked by this error.

(OTOH, it's possible we're completely misunderstanding the intent/expected usage pattern here.)

willbeason:
This fixes https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/issues/137

I _have_ tested this change in our operator, and it fixes our issue.

atoato88:
It's FYI.
I've checked this issue and [your PR](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/138).
In the same time, I've created [this PR](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/128), In my PR also fixes some (maybe) bugs in current `direct.go`.

When I checked your issue in my PR code-set, CRD and CR are created in success.
If you have a time, could you check my PR code-set (by cherry-pick)?

willbeason:
Sure, I'll be able to test your PR tomorrow. Thank you!

PR #138, opened 2020-12-10
Don't bail out if applying CRD + CR in manifest

willbeason:
The kubectl library upgrade for 1.18 now checks if all types are
available in applied manifests in res.Infos(). Callers should try to
apply everything returned if non-nil rather than immediately exiting on
an error.

This CL attempts to apply even if res.Infos() encounters an error. If
both .Apply() and res.Infos() return an error, returns the error from
.Apply().

I tried running the smoke test for this PR, but the smoke test seems broken: https://gist.github.com/willbeason/e3e35dafd7c982b08a3bf365387e55c4


willbeason:
This fixes #137

I have tested this change in our operator, and it fixes our issue.

PR #145, opened 2021-01-14
Embed Applier interface in kubectlClient interface 

darkowlzz:
Rename Applier interface method from `Applyx` to `Apply` and embed
the Applier interface in the `kubectlClient` interface.

### DIFF
--- a/pkg/patterns/declarative/reconciler.go
+++ b/pkg/patterns/declarative/reconciler.go
@@ -65,7 +65,7 @@ type Reconciler struct {
 }
 
 type kubectlClient interface {
-	Apply(ctx context.Context, namespace string, manifest string, validate bool, args ...string) error
+	applier.Applier
 }
 
 type DeclarativeObject interface {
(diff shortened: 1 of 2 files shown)


atoato88:
/lgtm
Nice catch!! and because of this, we can use this interface efficient.

atoato88:
@darkowlzz 
It's good to assign the approver to this PR for smooth merging process.
Please refer comments above from k8s-ci-bot in this PR.

darkowlzz:
Thanks @atoato88 .

/assign @mengqiy

pkg/patterns/declarative/reconciler.go
```diff
@@ -65,7 +65,7 @@ type Reconciler struct {
 }
 
 type kubectlClient interface {
-	Apply(ctx context.Context, namespace string, manifest string, validate bool, args ...string) error
+	applier.Applier
```
justinsb, line 68: This seems reasonable.  I do wonder if there's any reason to keep the `kubectlClient` interface, if we have the exported applier.Applier interface...

justinsb:
Thanks @darkowlzz 

/approve
/lgtm

pkg/patterns/declarative/reconciler.go
atoato88: Agree with your comment, we can use `Applier` interface instead of `kubectlClient` interface.

PR #164, opened 2021-06-17
Unwrap error when checking for not found.

briantkennedy:
This has been logging erroneous not found errors since the returned
error wraps the not found error.


pkg/patterns/declarative/reconciler.go
```diff
@@ -201,7 +201,7 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 	for _, obj := range objects.Items {
 
 		unstruct, err := GetObjectFromCluster(obj, r)
-		if err != nil && !apierrors.IsNotFound(err) {
+		if err != nil && !apierrors.IsNotFound(errors.Unwrap(err)) {
```
droot, line 204: Just a note (not scoped for this PR) that we should probably move to using new helpers available in the errors package. such as `errors.Is(...)`

droot (approved):

PR #167, opened 2021-08-16
Use labels helper to build label selector

justinsb:
A little more compact, and probably more correct.


atoato88:
/lgtm
I think It's more correct too. :smile:

PR #160, opened 2021-06-15
bump k8s to 1.21

varshaprasad96:
Co-authored-by: atoato88 <akihito-inou@nec.com>

**What this PR does / why we need it**:
This PR bumps controller runtime to 0.9.0 and k8s dependencies to 1.21.

PR #169, opened 2021-08-17
bump k8s to 1.22

atoato88:

This PR bumps k8s modules to 1.22.0

If this PR is too fast to merge because of other repository doesn't bump modules to 1.22, please add `/hold` tag.

**Update:**
I close this PR and will create other PR once controller-runtime release-0.10 tag is released.


atoato88:
Oh, should I update controller-runtime version in this PR?

justinsb:
Thanks for doing this @atoato88 ... I do think you're right in that we should try to follow the same k8s version as the controller-runtime version we're using.  My reasoning is that because client-go doesn't use true go module versioning, if there was a breaking change between client-go 1.21 and 1.22, and we were using 1.22 but controller-runtime was using 1.21, we might end up where the code can't be compiled.

It does look like controller-runtime was recently updated (in https://github.com/kubernetes-sigs/controller-runtime/pull/1626), but it also looks like that version hasn't yet been tagged.

I suggest once the version of controller-runtime is tagged, that we merge this change and update controller-runtime at the same time.  We could update controller-runtime now by pinning to a SHA, but I would prefer that we consume released/tagged versions of controller-runtime.

That said, if there's a particular feature in k8s 1.22 that you want, we could absolutely merge this with controller-runtime pinned to a SHA.  Is there?

I also suggest that we should start thinking about versioning and tagging in this repository also.  I think it is probably easiest to align with controller runtime. which describes their strategy here: https://github.com/kubernetes-sigs/controller-runtime/blob/master/VERSIONING.md  But in short I suggest we maintain a release-0.9 branch that works with controller-runtime-0.9, and in future a release-0.10 branch that works with controller-runtime 0.10.

If we want to do that, the big upside of doing that would be that master would probably be the 0.10 branch before it is released, which *would* use a SHA reference to the controller-runtime master branch.

That's a lot of words to say I think (1) we should think about release branches here and (2) if we are going to do release branches, we can merge this PR (but you should also update controller-runtime to the latest from the master branch in this PR).

atoato88:
> Thanks for doing this @atoato88 ... I do think you're right in that we should try to follow the same k8s version as the controller-runtime version we're using.  My reasoning is that because client-go doesn't use true go module versioning, if there was a breaking change between client-go 1.21 and 1.22, and we were using 1.22 but controller-runtime was using 1.21, we might end up where the code can't be compiled.
> It does look like controller-runtime was recently updated (in https://github.com/kubernetes-sigs/controller-runtime/pull/1626), but it also looks like that version hasn't yet been tagged.

Yes I agree with you and my PR was too fast timing. 😅 

> I suggest once the version of controller-runtime is tagged, that we merge this change and update controller-runtime at the same time.  However, we could update controller-runtime now by pinning to a SHA...

Yes, I think that we should update modules for this repository once controller-runtime is tagged.

As of now, should I close this PR and create other PR on that time? or hold this PR temporary?

justinsb:
> As of now, should I close this PR and create other PR on that time? or hold this PR temporary?

Actually I think if you want to update this PR to pull in the latest controller-runtime on the main/master branch (which uses k8s 1.22), then I think we can merge this.  If we agree that the versioning procedure I outline above makes sense,  our master branch will become release-0.10 once controller-runtime tags _their_ release-0.10.

We can also take a release-0.9 branch from immediately before this PR merges, if we decide we need that.  But I think just aiming for a solid release-0.10 (and then tagging 0.10.0 etc from it) is probably better.

atoato88:
> Actually I think if you want to update this PR to pull in the latest controller-runtime on the main/master branch (which uses k8s 1.22), then I think we can merge this. If we agree that the versioning procedure I outline above makes sense, our master branch will become release-0.10 once controller-runtime tags _their_ release-0.10.

The motivation that I create this PR is catch up with latest controller-runtime which is tagged.
Previously, [some PR](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/160) bumping some modules was along with same policy, I think. 

I agree with your versioning procedure comment.
So, I close this PR and create other PR once controller-runtime release-0.10 tag is released.

atoato88:
/close
Thank you to discussion.
I close this PR and will create other PR once controller-runtime release-0.10 tag is released.

varshaprasad96:
@atoato88 controller-runtime released [0.10.0](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.10.0) just yesterday. Can we reopen the PR?

varshaprasad96 (commented):
As the other projects have moved to go 1.16, can we also update the go version in declarative pattern? Looks like we are [still](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/0e495569c3aa05cc2c45c43eb76ee08c9991fa29/go.mod#L3) on 1.15

atoato88:
/reopen
Latest controller-runtime which taged `v0.10.0` is released.
I reopen this PR.

atoato88:
/retest

#175 is merged, I retest on this PR.

atoato88:
Oh, it isn't prow job..
I've rebased now. :sweat_smile:

estroz (approved):
/lgtm
/approve

PR #173, opened 2021-08-28
Fix namespace setting with direct applier

justinsb:
The direct-applier was not setting the namespace correctly; we reuse
the SetNamespace visitor to do so.

### DIFF
--- a/pkg/patterns/declarative/pkg/applier/direct.go
+++ b/pkg/patterns/declarative/pkg/applier/direct.go
@@ -2,48 +2,71 @@ package applier
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/printers"
 	"k8s.io/cli-runtime/pkg/resource"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
 	"k8s.io/kubectl/pkg/cmd/apply"
 	cmdDelete "k8s.io/kubectl/pkg/cmd/delete"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/patterns/declarative/pkg/applier/direct.go
```diff
@@ -32,16 +38,29 @@ func (d *DirectApplier) Apply(ctx context.Context,
...
 	}
-	restClient := genericclioptions.NewConfigFlags(true).WithDeprecatedPasswordFlag()
 	ioReader := strings.NewReader(manifest)
 
-	b := resource.NewBuilder(restClient)
+	restClientGetter := &staticRESTClientGetter{
+		RESTMapper: restMapper,
+		RESTConfig: restConfig,
+	}
+	b := resource.NewBuilder(restClientGetter)
```
atoato88, line 47: `Apply` has `validate` arg also, so we should add manifest validation if `validate` is `true` I think.

```suggestion
	b := resource.NewBuilder(restClientGetter)

	if validate {
		v, err := cmdutil.NewFactory(&genericclioptions.ConfigFlags{}).Validator(true)
		if err != nil {
			return err
		}
		b.Schema(v)
	}
```

I refered for initializing validator from [here](https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/apply/apply.go#L264).

atoato88:
Codes for adding Namespace is LGTM.
And I commented on manifest validation above.

pkg/patterns/declarative/pkg/applier/direct.go
justinsb: Thanks - merging.  One challenge that I'm trying to avoid is that when we build a new factory, we often have to repeat e.g. discovery against the server.

Maybe we should introduce an options struct here (1) to simplify the arguments and (2) so our API can be stable.

atoato88: Thank you to merging my suggestion.
I agree with your comment.
arguments as option is better solution, I think also.

PR #184, opened 2021-10-22
Skip setting owner refs for objects that don't support it

### LINKED ISSUE #176: Latest `kubebuilder-declarative-pattern` version stuck on endless reconcile loop for guestbook-operator
**What happened**:
Following the instructions of the latest version of the [walkthrough](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/tree/master/docs/addon/walkthrough#readme) results in a broken state, the operator enters an endless loop of create/delete of the resources defined in the `channels/packages/guestbook/0.1.0` folder. So the package is never in a `Ready` state.

I expected that the package would enter a stable state where it is `Ready`

1) Clone https://github.com/asilverman/guestbook-operator
2) Create a vanilla kind cluster with k8s version 1.12.1 (latest)
From the `guestbook-operator` folder, run (in sequence):
3) `make generate`
4) `make install`
5) `kubectl apply -n kube-system -f config/samples/addons_v1alpha1_guestbook.yaml`
6) `make run`

Inspect the logs of the operator and notice how it enters an endless loop

**Environment**:
```
clientVersion:
  buildDate: "2020-08-13T16:12:48Z"
  compiler: gc
  gitCommit: 9f2892aab98fe339f3bd70e3c470144299398ace
  gitTreeState: clean
  gitVersion: v1.18.8
  goVersion: go1.13.15
  major: "1"
  minor: "18"
  platform: linux/amd64
serverVersion:
  buildDate: "2021-05-21T23:01:33Z"
  compiler: gc
  gitCommit: 5e58841cce77d4bc13713ad2b91fa0d961e69192
  gitTreeState: clean
  gitVersion: v1.21.1
  goVersion: go1.16.4
  major: "1"
  minor: "21"
  platform: linux/amd64
```
- OS (e.g: `cat /etc/os-release`):
```
NAME="Ubuntu"
VERSION="20.04.1 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.1 LTS"
VERSION_ID="20.04"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```
- Kernel (e.g. `uname -a`):
```
Linux OFFICE-DT 5.10.16.3-microsoft-standard-WSL2 #1 SMP Fri Apr 2 22:23:49 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```
- Go version (e.g. `go version`): `go version go1.16.5 linux/amd64`

- Others:

tomasaschan:

Before this fix, a declarative controller using `declarative.WithOwner(declarative.SourceAsOwner)` will end up in an infinite reconciliation loop if any of the manifests in the package do not support having the declarative object as its owner.

For example, consider a _namespaced_ declarative object `foo/foo` of type `Foo`, installing a package that adds a few objects (say, two `Bar`s) in the same namespace plus another object (say, a `Baz`) in a fixed namespace `qux`. With the old behavior of this reconciler, the following happens:

1. The reconciler creates all objects in the package, setting owner references to an object with the same group, version, kind and name as the declarative object, and the same namespace as each respective object from the package. In other words, we get two `Bar`s in `foo` owned by `foo/foo`, and a `Baz` in `qux` owned by `qux/foo`.

2. Since `qux/foo` does not exist, the API server decides that it must have been deleted, and that means `qux/baz` should also be deleted (its owner no longer exists).

3. The reconciler is notified, and determines that `qux/baz` is missing, re-adding it again with an owner reference to `qux/foo`.

4. Go back to step 2.

With the change in this PR, the reconciler will instead skip setting an owner reference on `qux/foo`, avoiding the infinite loop at the cost of `qux/foo` now being a "dangling" object; if/when the declarative object `foo/foo` is deleted, `qux/foo` will be left untouched in the cluster.

Fixes #176

### DIFF
--- a/pkg/patterns/declarative/reconciler.go
+++ b/pkg/patterns/declarative/reconciler.go
@@ -493,9 +493,12 @@ func (r *Reconciler) injectOwnerRef(ctx context.Context, instance DeclarativeObj
 			continue
 		}
 
-		// TODO, error/skip if:
-		// - owner is namespaced and o is not
-		// - owner is in a different namespace than o
+		if owner.GetNamespace() != "" && owner.GetNamespace() != o.Namespace {
+			// a namespaced object can only own objects within the same namespace, not objects in other namespaces or cluster-scoped objects
+			// for any other combination, skip setting owner reference here, to allow declarative.SourceAsOwner to be used for the
+			// subset of objects that make up a supported combination
+			continue
+		}
 
 		ownerRefs := []interface{}{
 			map[string]interface{}{


pkg/patterns/declarative/reconciler.go
```diff
@@ -493,9 +493,12 @@ func (r *Reconciler) injectOwnerRef(ctx context.Context, instance DeclarativeObj
 			continue
 		}
 
-		// TODO, error/skip if:
-		// - owner is namespaced and o is not
-		// - owner is in a different namespace than o
+		if owner.GetNamespace() != "" && owner.GetNamespace() != o.Namespace {
+			// a namespaced object can only own objects within the same namespace, not objects in other namespaces or cluster-scoped objects
+			// for any other combination, skip setting owner reference here, to allow declarative.SourceAsOwner to be used for the
+			// subset of objects that make up a supported combination
```
estroz, line 499: This condition should be logged like above.

justinsb: I agree we should alert, but I've been thinking about who we are alerting for.  I think the answer is that we are alerting for the addon operator developer, but not the addon operator user.  There's actually a whole set of these, particularly as we improve the functionality (e.g. reuse of the RESTMapper), which include recommended but not required code changes.

I think I'm going to merge this PR as-is therefore, but then try to implement a "developer notification" mode.

tomasaschan: @justinsb I'm happy to amend this PR with logging if you'd like, but we currently have a rather ugly workaround with a custom `declarative.WithObjectTransform` that would be nice to get rid of, so if merging this and then adding new logging mechanism can be done in separate step's I'll be happy about it :)

PR #185, opened 2021-10-29
Use io and os in favor of ioutil

rajansandeep:
Cleanup deprecated ioutil and use io and os instead

Fixes #

**Additional documentation**:


justinsb:
Thanks for keeping us up to date!

/approve
/lgtm

PR #186, opened 2021-11-13
Improve watch event handling

justinsb:
Exit if we get an error notification.

Check for duplicate messages and suppress them, as they are all going
into the same channel.  This avoids renotification when the watch
connection is dropped.


atoato88:
/lgtm
Thank you to create!! :smile:

PR #187, opened 2021-11-13
Update go.mod to match controller-runtime 0.10.3

justinsb:
This should enable us up to tag 0.10.3, to match controller-runtime.


atoato88:
/lgtm
It's good to catch up with controller-runtime. :smile:

Issue #193, opened 2021-11-24
panic: concurrent map writes in watch/dynamic.go


tomasaschan:
We're seeing a `panic` on [`dw.lastRV[key] = rv`](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/7dcbbb6b50c8044fafa297997930d151152cde2a/pkg/patterns/declarative/pkg/watch/dynamic.go#L150) with the following stack trace:
```
"concurrent map writes"
Stack:
	2  0x000000000309d871 in sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative/pkg/watch.(*dynamicWatch).watchUntilClosed
	    at /home/tomasl/go/pkg/mod/ghe.spotify.net/declarative-infra/kubebuilder-declarative-pattern@v0.0.4/pkg/patterns/declarative/pkg/watch/dynamic.go:150
	3  0x000000000309c71a in sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative/pkg/watch.(*dynamicWatch).Add.func1
	    at /home/tomasl/go/pkg/mod/ghe.spotify.net/declarative-infra/kubebuilder-declarative-pattern@v0.0.4/pkg/patterns/declarative/pkg/watch/dynamic.go:86
```

It seems to happen every time our operator starts, but for different keys every time. I don't grok what goes on under the hood well enough to understand what different paths could be updating this simultaneously; what am I looking for in troubleshooting this? Or is it just a case of having too many resources? (This operator is looking at 1000+ instances of the CR, which each creates a couple of `RoleBinding`s and an [`IAMPartialPolicy`](https://cloud.google.com/config-connector/docs/reference/resource-docs/iam/iampartialpolicy) in the namespace where the CR is, plus a `RoleBinding` in a common namespace.)

If there's anything I can help out with to contribute a workaround/fix here, please let me know.

justinsb:
Ooops - thanks for reporting!

This should be a relatively simple fix - I forgot a mutex to guard lastRW.  However, the reason I forgot that mutex is that I thought that there should only be a single goroutine, but in fact there can be several if we are watching multiple types.  And because of that, lastRW should also be keyed by kind (or each kind needs its own map).

I can send a fix in an hour or two...  Adding a mutex is probably sufficient to avoid the crashes, but we/I should fix the map keying also.

tomasaschan:
Thanks!

[...]

PR #195, opened 2021-11-25
dynamic watch: scope lastRV to each watch

justinsb:
This avoids the race when we were (mistakenly) sharing the lastRV map.

Issue #193


atoato88:
/lgtm
I think also this PR fix race condition because that each kind has each lastRV as [said here](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/issues/193#issuecomment-977982233).

PR #197, opened 2022-01-06
refactory updating status mechanism and add custom private registry logic support

xuzhenglun:
**What this PR does / why we need it**:
This Pull Request includes two features:
1. refractory mechanism of updating status:
    a. allow status `Reconciled` method known if `error` happened, so that user could be able to submit an error to CR status field or launch an Event;
    b. maybe it's related with #190 
2. add the ability to custom private registry logic:
    for now, It always replaces the registry endpoint with a namespace (`placeholder/ns/image_name` --> `private_registry/image_name` ). But I want preserve namespace unchanged, (`placeholder/ns/image_name` --> `private_registry/ns/image_name` ). Compare with adding an new `ImageRegistryTransform`, leave `applyPrivateRegistryToImage` func customiable is a better way?

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Additional documentation**:

Issue #190, opened 2021-11-18
If an object transform fails, status is never updated


tomasaschan:
**What happened**:
Consider a declarative reconciler set up like this:
```go
func (r *MyReconciler) setupReconciler(mgr ctrl.Manager) (declarative.LabelMaker, error) {
    // ...
    
    return watchLabels, r.Reconciler.Init(mgr, &myapiv1.MyResource{},
        // ...
        declarative.WithStatus(status.NewBasic(mgr.GetClient()),
        declarative.WithObjectTransform(myTransform),
    )
}

func myTransform(ctx context.Context, object declarative.DeclarativeObject, objects *manifest.Objects) error {
    return errors.New("demonstrate the issue")
}
```

**What you expected to happen**:
When reconciling this resource, I would expect the status to be updated with `.status.healthy: false`, and at least the error message, perhaps even more detail, exposed in `.status.errors` and/or `.status.phase`. Instead, the reconciliation seems to crash completely, requeue itself and not update anything. This leaves the resource hanging without a status, and with no feedback to the user other than the fact that it doesn't have a status...

**How to reproduce it (as minimally and precisely as possible)**:
1. Define a declarative controller and resource
2. Add a simple transform function, e.g. the one in the example above, that returns an `error`
3. Observe the reconciliation behavior

**Anything else we need to know?**:
I know there's a comment on `NewBasic` that it's been superseeded by `NewKstatusCheck`, but I haven't been able to get the latter to actually update the status even in the happy-path case.

**Environment**:
- Kubernetes version (use `kubectl version`):
  ```
  λ kubectl version
  Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.4", GitCommit:"3cce4a82b44f032d0cd1a1790e6d2f5a55d20aae", GitTreeState:"clean", BuildDate:"2021-08-11T18:16:05Z", GoVersion:"go1.16.7", Compiler:"gc", Platform:"linux/amd64"}
  Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-21T23:01:33Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
  ```
- OS (e.g: `cat /etc/os-release`):
  ```
  λ cat /etc/os-release 
[...]

PR #199, opened 2022-01-17
refactory updating status mechanism, status func could handle error

xuzhenglun:

related with #197: 

1. refractory mechanism of updating status:
    a. allow status `Reconciled` method known if `error` happened, so that user could be able to submit an error to CR status field or launch an Event;
    b. maybe it's related with #190 

Fixes #

**Additional documentation**:

### DIFF
--- a/pkg/patterns/addon/pkg/status/aggregate.go
+++ b/pkg/patterns/addon/pkg/status/aggregate.go
@@ -27,6 +27,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+
 	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative"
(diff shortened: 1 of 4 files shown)


atoato88:
/lgtm
I think contents of this PR don't have any side effects, so It's OK.
`Reconciled` function doesn't use `error` object on 4th args at [here](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/199/files#diff-f4f154817e266917cf48c21f464818b16798c1d4e1170cee1af9804481f786f0R48) and [here](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/199/files#diff-eb527f90ab4b8fa744a6601d127c711bdde2d3cd9956532a736738203a4fc94fR26).
But on other PR, we can use this object to write more information on status, I think.

atoato88:
/assign @justinsb 
Could you review this?

xuzhenglun:
@justinsb please take a look when you are convenient.

atoato88:
/approve
This is good update for this repository and confirmed that tests past successfully.
I approve this.

PR #205, opened 2022-03-02
Add unit test for loaders/types

atoato88:
**What this PR does / why we need it**:

This PR adds more unit tests for `pkg/patterns/addon/pkg/loaders/types.go`.
As of now, this repository should have more unit tests, because of codebase quality and readability for logics which already exists.

**Which issue(s) this PR fixes**:

**Special notes for your reviewer**:

**Additional documentation**:

PR #207, opened 2022-03-07
Add unit test for loaders/fs

atoato88:

This PR adds more unit tests for `pkg/patterns/addon/pkg/loaders/fs.go`.

NONE

NONE

**Additional documentation**:
related to: #205

### DIFF
--- a/pkg/patterns/addon/pkg/loaders/fs_test.go
+++ b/pkg/patterns/addon/pkg/loaders/fs_test.go
@@ -1,0 +1,209 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package loaders
+
+import (
+	"context"
+	"reflect"
+	"testing"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	addonv1alpha1 "sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/addon/pkg/apis/v1alpha1"
+)
+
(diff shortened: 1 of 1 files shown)


pkg/patterns/addon/pkg/loaders/fs_test.go
```diff
@@ -0,0 +1,209 @@
...
+			expected: &HTTPRepository{},
+		},
+		{
+			name:     "http pattern",
+			channel:  "https://example.com",
+			expected: &HTTPRepository{},
+		},
+		{
+			name:     "git pattern",
+			channel:  "git//example.com",
```
atoato88, line 52: What pattern implemented logic supposes at?
https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/fea7e5cc701290589ec20ef4d9c0629d08b5307d/pkg/patterns/addon/pkg/loaders/fs.go#L50

I don't know the pattern about `git//`(w/o colon). something special pattern for git? :thinking:

justinsb: Good question.  I think it's a typo and it should be `.git//`

Change introduced here without explanation: https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/commit/4bf81ecfa10b81a0c8fd092b396d38f802a480e0

The `//` syntax is actually from kustomize: https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md  

> The URL format is an HTTPS or SSH git clone URL with an optional directory and some query string parameters. The directory is specified by appending a // after the repo URL. The following query string parameters can also be specified:


So it's actually looking for something like `https://github.com/kubernetes/kubernetes//subdir1/subsubdir`, it's not actually looking for `git://` because `git` doesn't itself have a transport (it leverages https or SSH)

justinsb: Looks like I actually made an interesting typo here .... it's looking for ` `https://github.com/kubernetes/kubernetes.git//subdir1/subsubdir`, but kustomize also accepts `https://github.com/kubernetes/kubernetes//subdir1/subsubdir`.  I think the problem we have is that `//` is more likely to be a typo / mistake the https url construction (maybe), and so I don't think we should assume that every https url with two slashes is in fact a git url.

The `.git//` check is a compromise IIUC, because github accepts `https://github.com/org/repo.git` as a synonym for `https://github.com/org/repo`.

justinsb:
Interesting observation on the git urls, but this PR looks good to me.

I think if we wanted to change `git//` to `.git//` we should probably bump the version; better versioning is something I've been wanting to do for a while, so maybe this is the time!

/approve
/lgtm

PR #208, opened 2022-03-10
Add unit test for declarative/labels

atoato88:
This PR adds unit tests for `pkg/patterns/declarative/labels_test.go`.

NONE

NONE

**Additional documentation**:
I would like to add unit tests for package `pkg/patterns/declarative` because of this package has many manipuration logics of this repository.

### DIFF
--- a/pkg/patterns/declarative/labels_test.go
+++ b/pkg/patterns/declarative/labels_test.go
@@ -1,0 +1,225 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package declarative
+
+import (
+	"context"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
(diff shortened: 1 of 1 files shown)


pkg/patterns/declarative/labels_test.go
```diff
@@ -0,0 +1,225 @@
...
+  strategy: {}
+  template:
+    metadata:
+      labels:
+        app: test-app
+    spec:
+      containers:
+      - image: busybox
+        name: busybox`
+	var testcast = []struct {
```
justinsb, line 57: Nit: typo - I think this should be testcases (?)

justinsb (approved):

justinsb:
Thanks @atoato88 ... we could also consider creating some golden tests (e.g. https://ieftimov.com/posts/testing-in-go-golden-files/ ) that test whole operator functionality, and then do a few tests with the various options.  But this lgtm!

pkg/patterns/declarative/labels_test.go
atoato88: oh.. correct.
Thank you to comment :smile:

atoato88:
Thank you to approve PR and give idea about testing!! :smile: 
I'll check the URL.

atoato88:
@justinsb 
Thank you to let me know the article about golden file test.
I agree with this plan, so should we implement operator (maybe dummy operator?) for testing functionalities for this repo ?
Or, we should be versioning this repo before this?

justinsb:
I created the release-0.11 branch as a first step towards versioning!  But creating the branch doesn't really commit us to anything :-)

I don't think tests should impact the release versioning (unless they find problems!).  I think we should use versioning for breaking changes, and ideally we would backport fixes at least to the current released version while we're developing the next version.  So if we start thinking of master as the upcoming 0.12.0 release, then we should backport any bugfixes to the release-0.11 branch; we would expect operators that want stability to use release-0.11 until we release release-0.12.

Because it's a library, "releasing" just means tagging I think.

And yes, in terms of the testing I was thinking we should create a test operator and maybe expose a few hooks so that it's easy to create derived tests testing different configurations.  And we should run it in our e2e against the k8s testclient.  Similar to the work you did to create golden tests like [this one](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/96a66dfbc7d70af79ba10c69b6a786b1229708b6/examples/guestbook-operator/controllers/guestbook_controller_test.go) - I'm wondering if we can use that approach more widely and for more configurations.  But it's just a thought, if you want to explore it that's wonderful, otherwise hopefully I'll be able to explore it!

atoato88:
Thank you to detailed comment.
Yes, I agree with your plan, and we need test operator like [`guestbook-operator`][1] and [tests for that][2].
It's good to create test operator at first, and expand cover range about test functionalities, I think.

[1]: https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/examples/guestbook-operator/controllers/guestbook_controller.go
[2]: https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/examples/guestbook-operator/controllers/guestbook_controller_test.go

PR #210, opened 2022-03-24
don't remove objects with ignore annotation when prune is enabled

### LINKED ISSUE #209: pruning and resource ignore annotation is conflict
**What happened**:
1. Controller is enabled `prune`, so that resources can be deleted automatically after they are not in manifests.  
2. After annotating a resource with `addons.k8s.io/ignore=true`, this resource will be deleted immediately and re-apply.

resources with `addons.k8s.io/ignore=true` will be ignored after applying, and will not be pruned.

1. now pruning is broken because of https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/issues/201#issuecomment-1077341195
2. after workarounds above issue, run `kubectl annotate foo addons.k8s.io/ignore=true`, this `foo` will be removed and then recreated. 

**Environment**:
kubernetes version is `1.20.12`

- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Go version (e.g. `go version`):
- Others:

xuzhenglun:
fix #209 

**Additional documentation**:

### DIFF
--- a/pkg/patterns/declarative/reconciler.go
+++ b/pkg/patterns/declarative/reconciler.go
@@ -230,6 +230,25 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 			if _, ok := annotations["addons.k8s.io/ignore"]; ok {
 				log.WithValues("kind", obj.Kind).WithValues("name", obj.GetName()).Info("Found ignore annotation on object, " +
 					"skipping object")
+
+				// remove labels, avoiding be removed when kubectl.Apply is called with flag `prune`
+				needUpdate := false
+				{ // remove labels from object
+					labels := unstruct.GetLabels()
+					for k := range r.options.labelMaker(ctx, instance) {
+						if _, ok := labels[k]; ok {
+							needUpdate = true
+						}
+						delete(labels, k)
+					}
+					unstruct.SetLabels(labels)
+				}
+				if needUpdate { // update object if it's necessary
+					if _, err := UpdateObjectFromCluster(unstruct, r); err != nil {
+						return objects, fmt.Errorf("remove label for resource with 'addons.k8s.io/ignore' failed, %w", err)
+					}
+				}
+
 				continue
 			}
 		}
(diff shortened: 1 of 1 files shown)


xuzhenglun:
@atoato88

remove labels if `ignore` annotation is applied, does this solution ok for you?

pkg/patterns/declarative/reconciler.go
```diff
@@ -230,6 +230,25 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
+						if _, ok := labels[k]; ok {
+							needUpdate = true
+						}
+						delete(labels, k)
+					}
+					unstruct.SetLabels(labels)
+				}
+				if needUpdate { // update object if it's necessary
+					if _, err := UpdateObjectFromCluster(unstruct, r); err != nil {
+						return objects, fmt.Errorf("remove label for resource with 'addons.k8s.io/ignor' failed, %w", err)
```
atoato88, line 248:
```suggestion
						return objects, fmt.Errorf("remove label for resource with 'addons.k8s.io/ignore' failed, %w", err)
```
maybe typo?

atoato88:
@xuzhenglun 
IMHO, labels set in `r.options.labelMaker` are quite important on some situations.
e.g. Filter objects by those labels, ...

@justinsb 
Any ideas which don't delete those labels?

xuzhenglun:
> @xuzhenglun IMHO, labels set in `r.options.labelMaker` are quite important on some situations. e.g. Filter objects by those labels, ...
> 
> @justinsb Any ideas which don't delete those labels?

@atoato88

Or maybe we can add another label, such as `addon.k8s.io/managed=true` to all addon objects.  

Later, In prune progress, we can append a selector: `addon.k8s.io/managed!=true`. The corresponding `kubectl` command  could be `kubectl apply --prune --selector=<user-labels>,addon.k8s.io/managed!=true`

Both of the two situations are ok for me. After deciding which one we should choose, then I'm willing to update this PR.

pkg/patterns/declarative/reconciler.go
xuzhenglun: oh, indeed. I will update this PR until we find a right way with this https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/210#issuecomment-1078749278

atoato88:
Thank you to comment your ideas.
I think 2nd one is more reasonable for me;
> Or maybe we can put a label addons.k8s.io/ignore on the object which contains the annotation addons.k8s.io/ignore.


But it's a breaking change from current codebase, so we should decide this with other approvers.

@justinsb 
WDYT?

pkg/patterns/declarative/reconciler.go
```diff
@@ -230,6 +230,25 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
+				needUpdate := false
+				{ // remove labels from object
+					labels := unstruct.GetLabels()
+					for k := range r.options.labelMaker(ctx, instance) {
+						if _, ok := labels[k]; ok {
+							needUpdate = true
+						}
+						delete(labels, k)
+					}
+					unstruct.SetLabels(labels)
```
JeffLuoo, line 244: If no reconcile is triggered, how those labels get set back?

PR #211, opened 2022-03-24
running kubectl apply in higher level

### LINKED ISSUE #201: No pruning with NewDirectApplier
**What happened**:
The `kubectl --prune` flag is set as `extraOpts` in [`reconciler.go`](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/pkg/patterns/declarative/reconciler.go#L236).

However, since the `NewDirectApplier` is used (https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/101), the prune option is not propagated to the `applyOpts.Prune` boolean of [`direct.go`](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/pkg/patterns/declarative/pkg/applier/direct.go#L80).

The prune option to be propagated so that pruning happens as expected.

Putting a breakpoint on [`direct.go#L80`](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/pkg/patterns/declarative/pkg/applier/direct.go#L80) clearly shows that the `Prune` flag is not set on the `applyOpts` object.

There currently is also no support for the `--prune-whitelist` flag, which we need for our use case.
Happy to create a PR for this, but I think it's better to solve this issue first.

Is there a specific reason for keeping the old ExecApplier around?
If not, I guess the `ApplierOptions` from the library could be directly replaced by [`ApplyOptions`](https://github.com/kubernetes/kubectl/blob/8d3e8f9f6868315eb7204b7d511fac0abdc13e5c/pkg/cmd/apply/apply.go#L75) from the `k8s.io./kubectl` module.

**Environment**:
```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.5", GitCommit:"aea7bbadd2fc0cd689de94a54e5b7b758869d691", GitTreeState:"clean", BuildDate:"2021-09-15T21:10:45Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-21T23:01:33Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
```
- OS (e.g: `cat /etc/os-release`): macOS Big Sure 11.6.2
- Kernel (e.g. `uname -a`): 
```
Darwin C02YK2Z2JHD2 20.6.0 Darwin Kernel Version 20.6.0: Wed Nov 10 22:23:07 PST 2021; root:xnu-7195.141.14~1/RELEASE_X86_64 x86_64 i386 MacBookPro15,2 Darwin
```
- Go version (e.g. `go version`): `go version go1.17.6 darwin/amd64`
- Others:

xuzhenglun:

fix #201

Fixes #

**Additional documentation**:

### DIFF
--- a/pkg/patterns/declarative/pkg/applier/direct.go
+++ b/pkg/patterns/declarative/pkg/applier/direct.go
@@ -1,20 +1,22 @@
 package applier
 
 import (
+	"bytes"
 	"context"
 	"fmt"
-	"os"
 	"strings"
 
+	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
-	"k8s.io/cli-runtime/pkg/printers"
 	"k8s.io/cli-runtime/pkg/resource"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
 	"k8s.io/kubectl/pkg/cmd/apply"
-	cmdDelete "k8s.io/kubectl/pkg/cmd/delete"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 type DirectApplier struct {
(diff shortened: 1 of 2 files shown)


xuzhenglun:
@atoato88 

please take a look at this PR too. thanks

pkg/patterns/declarative/pkg/applier/direct.go
```diff
@@ -121,3 +102,52 @@ func (s *staticRESTClientGetter) ToRESTMapper() (meta.RESTMapper, error) {
...
+
+func (namespaceStub) ClientConfig() (*rest.Config, error) {
+	panic("implement me")
+}
+
+func (namespaceStub) ConfigAccess() clientcmd.ConfigAccess {
+	panic("implement me")
+}
+
+func NewCmdApply(ioStreams genericclioptions.IOStreams) (*cobra.Command, *apply.ApplyOptions) {
```
atoato88, line 128: Did you copy some codes from [k/kubectl](https://github.com/kubernetes/kubectl/blob/v0.22.2/pkg/cmd/apply/apply.go#L163)?
IMHO, it's good not to copy, but use modules.
I think `DirectApplier` is implementend with something tricky way.

Recently, we started these topics related with `prune` functionality at #201.
We may have to have discussion at #201.

xuzhenglun: yes, It's copied from `k/kubectl`.  In the `NewCmdApply` function, `ApplyOptions` is never returned, so I didn't find a way to reuse that piece of code. 

In the previous implementation, some codes are copied from [Complete](https://github.com/kubernetes/kubectl/blob/920da0b7c6952cfe2cdbed4c65e9addc0f7be219/pkg/cmd/apply/apply.go#L209) too, with some customize modification, Which can be very easy to make mistakes. I agreed`DirectApplier` is implemented very trickly indeed.

xuzhenglun:
I believe that #217 fixes #201 too, and we can close this PR now.

jcanseco:
FYI we started to encounter an issue that seems to have been introduced in this PR. Wondering if someone can take a look or if there is a workaround we can use until a fix is out: https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/issues/224

jcanseco:
(Deleted reply; wrong thread)

PR #212, opened 2022-03-24
Introduce setters for name and namespace

justinsb:
This allows changing them - in particular setting the namespace.
Without this if the Namespace field was changed, it would not be
reflected in the underlying object.
    
This is also more similar to unstructured.Unstructured, which we are
modelling after.

Also proposing that we think of this as the first breaking change for 0.12.  We can continue to cherry-pick back to 0.11 though!

### DIFF
--- a/docs/releases/release-0.12.md
+++ b/docs/releases/release-0.12.md
@@ -1,3 +1,7 @@
 # Release notes for 0.12 release
 
 Note: the 0.12 release is pending.  This serves to accumulate breaking/important changes prior to release.
+
+* Name and Namespace are not longer exported by manifest.Object; use GetName() and GetNamespace() instead.
+  This change allows setting the values.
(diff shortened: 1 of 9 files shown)


atoato88:
CI test failed because we need update some place to use new getter.
I checked this codebase on my local, and create [commit on my fork][1] to fix test failure.
So please check that.

[1]: https://github.com/atoato88/kubebuilder-declarative-pattern/commit/88c8a89c13aa683c13c6660ae069f01557ae2d6b

atoato88:
Except for my comment above, it's LGTM :smile:

justinsb:
Thanks @atoato88 - I merged in your changes and marked you as a co-author :-)

atoato88:
/lgtm
Thank you to update!

PR #213, opened 2022-03-25
check object scope before fetching

xuzhenglun:

if `metadata.namespace` on Cluster-level resource(such as ClusterRole) is set, `kubectl apply` will not report an error and just ignore it. But when I submit via Controller, it failed and report an error like `resource xxx is not found`. 

After some time I dug, I realize that `namespace` is joined into the URL. So maybe we should check whether the resource is Cluster-level, and ignore `metadata.namespace` if it's positive.

Fixes #

**Additional documentation**:


xuzhenglun:
ping @atoato88

I rebased this PR and CI is passed.  Please leave an comment when you are convenient.

atoato88:
/lgtm
/approve
Thank you to rebase :smile:

PR #214, opened 2022-03-28
enhance handling prune for CR

xuzhenglun:

1. allow CR to decide whether to enable pruning resources;
2. allow CR to decide what kinds of resources should be pruned.

Fixes #

introduce 2 interfaces to let CR can decide whether and how to prune resources. When none of them are implemented, the behavior should be the same as before. So I think this is not a breaking change.

**Additional documentation**:

### DIFF
--- a/pkg/patterns/declarative/reconciler.go
+++ b/pkg/patterns/declarative/reconciler.go
@@ -74,6 +74,21 @@ type DeclarativeObject interface {
 	metav1.Object
 }
 
+// Pruner is a trait for addon CRDs that determines whether pruning behavior should be enabled for the current CR.
+// To enable this feature, it's necessary to enable WithApplyPrune. If WithApplyPrune is enabled but Pruner is not
+// implemented, Prune behavior is assumed by default.
+type Pruner interface {
+	Prune() bool
+}
+
+// PruneWhiteLister is a trait for addon CRDs that determines which kind of resources should be pruned. It's useful
+// when CR in installed by Addon and want to prune them automatically. The format of array item should be exactly like
+// <group>/<version>/<kind> (core group using 'core' indeed). For example: ["core/v1/ConfigMap", "batch/v1/Job"].
+// Notice: kubeadm has a built-in prune white list, and it will be ignored if this method is implemented.
+type PruneWhiteLister interface {
+	PruneWhiteList() []string
+}
+
 type ErrorResult struct {
 	Result reconcile.Result
 	Err    error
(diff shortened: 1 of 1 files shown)


xuzhenglun:
ping @atoato88

This PR enhances the ability of `prune`, and it doesn't relate with `DirectApplier`.  I would be very grateful if you could take a look at this PR.

atoato88:
I agree with this PR's enhancement about `prune-whitelist`.

I think `kubebuilder-declarative-pattern` has similar functions by [`WithApplyPrune`](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/pkg/patterns/declarative/options.go#L108-L113) already.
I think this framework chosed implement `With***` functions for manipurating, so it's prefer to update `WithApplyPrune` or add new `With***` functions if you would like to enhance `prune` functionality.
e.g. implement new `WithPruneWhitelist`?
WDYT?

xuzhenglun:
> agree

`With***` function works just fine when we want to build a specific addon operator.  However, I'm trying to implement a generic AddonManager for Cluster.  My user could choose to enable `pruning` by setting a flag in CR. Feature more, consisting of Addon could have CRD inside, so they should be able to configure the white list to determine what kind of resource could be pruned.

To achieve the above by using the `WithPrune` function, It's necessary to pass CR to the user's function, and it will change the function signature of `WithPrune` which can be a breaking change.  Secondly, framework users still need to implement methods like `Prune` or `PruneWhiteList` anyway, because those configurations are defined in CR.

WDYT?

pkg/patterns/declarative/reconciler.go
```diff
@@ -74,6 +74,14 @@ type DeclarativeObject interface {
 	metav1.Object
 }
 
+type Pruner interface {
```
atoato88, line 77: Could you add some explaination for this interface?

pkg/patterns/declarative/reconciler.go
```diff
@@ -74,6 +74,14 @@ type DeclarativeObject interface {
...
+	Prune() bool
+}
+
+type PruneWhiteLister interface {
```
atoato88, line 81: ditto.

atoato88:
Thank you to explain the context detailed.
I agree with your thinking, I think it's preferable for achieve your requirements.

BTW, I left review comments above, please check those.

pkg/patterns/declarative/reconciler.go
atoato88: And, `[]string` has specific format for `gvk`?
If so, please wirte that.

pkg/patterns/declarative/reconciler.go
xuzhenglun: An explanation of the interface has been appended.

xuzhenglun:
> Thank you to explain the context detailed. I agree with your thinking, I think it's preferable for achieve your requirements.
> 
> BTW, I left review comments above, please check those.

Thanks for your review. I append some comments in the source code and just updated this PR.

pkg/patterns/declarative/reconciler.go
atoato88: perfect! :smile:

atoato88:
/approve
Thank you to update quickly!

atoato88:
/approve cancel
oh, we should check test before approving.

xuzhenglun:
@atoato88 still needs an LGTM label to merge. Is something I need to do yet?

PR #215, opened 2022-03-31
Allow multiple resources in manifest files when using kustomize

llamallamaduck:
If there are multiple resources in a manifest, and `WithApplyKustomize` option is used, the resulting manifests to apply would only contain the last item from the manifest. This PR corrects this, and takes into account that a manifest could contain multiple resources.


atoato88:
/lgtm

Nice catch!
It's important fix... :smile:

PR #218, opened 2022-05-12
RESTMapper optimized for controllers

justinsb:
The default RESTMappers are very optimized around the kubectl use-case, where most of the time controllers only need to map GVK -> GVR to support an apply operation.

Start adding a RESTMapper optimized around this specific use-case.

### DIFF
--- a/pkg/restmapper/caching.go
+++ b/pkg/restmapper/caching.go
@@ -1,0 +1,115 @@
+package restmapper
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// cache is our cache of schema information.
+type cache struct {
+	mutex         sync.Mutex
+	groupVersions map[schema.GroupVersion]*cachedGroupVersion
+}
+
+// newCache is the constructor for a cache.
+func newCache() *cache {
+	return &cache{
+		groupVersions: make(map[schema.GroupVersion]*cachedGroupVersion),
+	}
+}
+
+// cachedGroupVersion caches (all) the resource information for a particular groupversion.
+type cachedGroupVersion struct {
+	gv    schema.GroupVersion
+	mutex sync.Mutex
+	kinds map[string]cachedGVR
+}
+
+// cachedGVR caches the information for a particular resource.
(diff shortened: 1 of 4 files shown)


pkg/restmapper/controllerrestmapper.go
```diff
@@ -0,0 +1,80 @@
...
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
+)
+
+// NewControllerRESTMapper is the constructor for a ControllerRESTMapper
+func NewControllerRESTMapper(cfg *rest.Config) (meta.RESTMapper, error) {
```
apelisse, line 13: I'm assuming if you want to reset the cache you just reinstantiate a new controller restmapper?

pkg/restmapper/caching.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	return &meta.RESTMapping{
+		Resource:         c.gv.WithResource(cached.Resource),
+		GroupVersionKind: c.gv.WithKind(kind),
+		Scope:            cached.Scope,
+	}, nil
+}
+
+// fetch returns the metadata, fetching it if not cached.
+func (c *cachedGroupVersion) fetch(discovery discovery.DiscoveryInterface) (map[string]cachedGVR, error) {
```
apelisse, line 73: I'm not sure I understand how this is different from the existing restmapper fundamentally? Except that you already have the groupversion so you don't need to target the `/api/` discovery root endpoint?

apelisse (commented):
I'm curious how much this could be solved in k/k directly, since that could more useful in general. Thanks!

justinsb:
> I'm curious how much this could be solved in k/k directly, since that could more useful in general. Thanks!

Yes - absolutely - we should use the subprojects... let's prove that it works in k-sigs, then promote it to k/k (e.g. client-go).  If we need something from e.g. the apiserver to improve things, we'll figure that out here!

pkg/restmapper/controllerrestmapper.go
justinsb: Yes, or it's a future enhancement!  For example, if we create a CRD, we might want to invalidate just the GV we created.

pkg/restmapper/caching.go
justinsb: So the current restmapper does discovery on every GV in the system.  So when we register more CRDs, every controller slows down.  Most controllers only work with a small subset of GVs, and so the goal is that they only load the ones they care about.  There's a small memory win, but it's really about much faster startup time (and avoiding the "thundering herd" effect).

apelisse:
Thanks,
/lgtm
/approve

PR #223, opened 2022-06-07
try apply yaml in best efforts

xuzhenglun:
1. When I apply CR with it's CRD, an error like `resource <Kind> is not found` will return. 
2. In fact, `kubectl` will try its best efforts to apply YAMLs. In the progress, maybe CRDs have not been established and an error could return. But in the next retry, all YAMLs can be applied successfully.

ref: https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/apply/apply.go#L438

Fixes #

**Additional documentation**:

### DIFF
--- a/pkg/patterns/declarative/pkg/applier/direct.go
+++ b/pkg/patterns/declarative/pkg/applier/direct.go
@@ -8,6 +8,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/printers"
 	"k8s.io/cli-runtime/pkg/resource"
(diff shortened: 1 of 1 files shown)


pkg/patterns/declarative/pkg/applier/direct.go
```diff
@@ -79,7 +85,7 @@ func (d *DirectApplier) Apply(ctx context.Context, opt ApplierOptions) error {
 	applyCmd := apply.NewCmdApply(baseName, f, ioStreams)
 	applyOpts, err := applyFlags.ToOptions(applyCmd, baseName, nil)
 	if err != nil {
-		return fmt.Errorf("error getting apply options: %w", err)
+		return utilerrors.NewAggregate(append(errs, fmt.Errorf("error from SetNamespace: %w", err)))
```
atoato88, line 88: Is message of `error from SetNamespace` correct?

pkg/patterns/declarative/pkg/applier/direct.go
```diff
@@ -106,7 +112,8 @@ func (d *DirectApplier) Apply(ctx context.Context, opt ApplierOptions) error {
 		IOStreams: ioStreams,
 	}
 
-	return applyOpts.Run()
+	err = applyOpts.Run()
+	return utilerrors.NewAggregate(append(errs, fmt.Errorf("error from SetNamespace: %w", err)))
```
atoato88, line 116: ditto.

atoato88 (commented):
Thank you to create PR.
It looks almost good but left comment.
Please check those.

pkg/patterns/declarative/pkg/applier/direct.go
xuzhenglun: ho, I just corrected it.

xuzhenglun:
@[atoato88](https://github.com/atoato88) Is anything else I need to do?

PR #228, opened 2022-07-11
Parsing: add more empty test cases

justinsb:
Add some test cases for the empty object cases, as these have been
problematic in the past.


atoato88:
/lgtm
Nice follow up!
It's great point :smile:

PR #229, opened 2022-07-13
Fix bug where Reconcile() returns nil err if status option is not nil

jcanseco:

This commit fixes an issue where `Reconcile()` can return a nil `err`,
even if an error occured during execution, as long as the `status`
option is not nil.

This bug effectively hid from callers of `Reconcile()` if an error
occurred.

Fixes #

**Additional documentation**:

This issue seems to have originated from https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/199.

### DIFF
--- a/pkg/patterns/declarative/reconciler.go
+++ b/pkg/patterns/declarative/reconciler.go
@@ -170,8 +170,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		}
 
 		if r.options.status != nil {
-			if err = r.options.status.Reconciled(ctx, instance, objects, err); err != nil {
-				log.Error(err, "failed to reconcile status")
+			if statusErr := r.options.status.Reconciled(ctx, instance, objects, err); statusErr != nil {
+				log.Error(statusErr, "failed to reconcile status")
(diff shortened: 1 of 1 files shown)


pkg/patterns/declarative/reconciler.go
```diff
@@ -170,8 +170,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 		}
 
 		if r.options.status != nil {
-			if err = r.options.status.Reconciled(ctx, instance, objects, err); err != nil {
-				log.Error(err, "failed to reconcile status")
+			if statusErr := r.options.status.Reconciled(ctx, instance, objects, err); statusErr != nil {
```
justinsb, line 173: _Maybe_ we should return an error, if `err` was nil going in to the defer block, but then we failed to update the status, so we retry again to update the status.  However, I think this fix is likely more important.

justinsb:
Thanks @jcanseco - nice fix!

/approve
/lgtm

atoato88:
Thank you to active contribution!
LGTM (I know it has been merged already :sweat_smile:

PR #233, opened 2022-07-27
Add unit test for direct applier

atoato88:
This PR adds unit tests for direct applier in `direct.go`.

`direct.go` doesn't have tests as far, we should add tests for prevent degrades.

For implimenting tests, this PR changes direct applier to have some delegation structures.
It is [the same manner with that in `exec.go`](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/2072cf081b2e512b687c6bf152466646df1ac2c9/pkg/patterns/declarative/pkg/applier/exec.go#L31-L51)

Related to [#201 comment](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/issues/201#issuecomment-1185585708)

None

**Additional documentation**:
None

### DIFF
--- a/pkg/patterns/declarative/pkg/applier/direct_test.go
+++ b/pkg/patterns/declarative/pkg/applier/direct_test.go
@@ -28,6 +28,8 @@ import (
 	"k8s.io/client-go/rest/fake"
 	"k8s.io/client-go/restmapper"
 	"k8s.io/kubectl/pkg/cmd/apply"
+	kubectltesting "k8s.io/kubectl/pkg/cmd/testing"
+	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/scheme"
 )
 
@@ -53,6 +55,10 @@ func (d *directApplierTestSite) NewBuilder(opt ApplierOptions) *resource.Builder
 		func() (restmapper.CategoryExpander, error) {
 			return resource.FakeCategoryExpander, nil
 		})
+}
+
+func (d *directApplierTestSite) NewFactory() cmdutil.Factory {
+	return kubectltesting.NewTestFactory()
 }
 
 func newDirectApplierTest(d *directApplierTestSite) Applier {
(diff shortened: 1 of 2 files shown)


atoato88:
/assign @justinsb 
Could you review this?

pkg/patterns/declarative/pkg/applier/direct_test.go
```diff
@@ -0,0 +1,194 @@
...
+				Prune: true,
+			},
+			expectCheckFunc: func(opt *apply.ApplyOptions) error {
+				if opt.Prune != true {
+					return fmt.Errorf("prune is not set")
+				}
+				return nil
+			},
+		},
+		//		{
```
atoato88, line 148: I found that current codebase [doesn't refer `--prune-whitelist`](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/2072cf081b2e512b687c6bf152466646df1ac2c9/pkg/patterns/declarative/pkg/applier/direct.go#L92-L102).
That's why I commented out here.

atoato88:
umm, [`test.sh`](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/hack/ci/test.sh) passed on my laptop, but failure on CI...
It looks like that can't use fake client? :thinking:

atoato88:
Current PR needs live k8s cluster, I think.
Is it possible that unit test on fake client without k8s cluster...?

atoato88:
/retitle [WIP] Add unit test for direct applier

pkg/patterns/declarative/pkg/applier/direct_test.go
```diff
@@ -0,0 +1,204 @@
...
+			expectCheckFunc: func(opt *apply.ApplyOptions) error {
+				if opt.Namespace == "test-namespace" {
+					return nil
+				} else {
+					return fmt.Errorf("namespace doesn't match to \"test-namespace\"")
+				}
+			},
+		},
+		// This test use fake test factory on "k8s.io/kubectl/pkg/cmd/testing", that's why this test for validate is commented out.
+		// opt.Validator is always set to validation.NullSchema instance by fake test factory, so we can't check if validator is set successfully.
```
atoato88, line 114: We should this test w/o live k8s cluster, so I use fake factory.
Unit test become to pass by using that, but we can't check validate options.. :thinking: 
Is there any good implementations?

atoato88:
/retitle Add unit test for direct applier
Now, ci test become to pass, I remove WIP.

pkg/patterns/declarative/pkg/applier/direct_test.go
justinsb: Sorry for delay!  I just uploaded a test with a fake kubeapiserver that is being developed across a few projects using copy-and-paste reuse while we figure it out!  (The projects being kOps, kpt and this one!)

I still want this PR as I agree we need more test coverage, but I think knowing that the other PR can cover the HTTP round trips makes not validating here OK.

Although actually, validation is generally moving to the server, so we might be OK to just discourage client-side validation anyway...

pkg/patterns/declarative/pkg/applier/direct_test.go
justinsb: Do you use prune-whitelist?  If not, I propose we make the direct applier strict, i.e. it should reject unknown options.  Thus it should error if someone passes `prune-whitelist`.

If you do use it, can we map it to the underlying library?

PR #238, opened 2022-08-29
update k8s to 1.25 and go to 1.19

everettraven:
Updates Kubernetes dependencies to 1.25 and Go to 1.19 to provide support for using `kubebuilder-declarative-pattern` with k8s 1.25

Fixes #

**Additional documentation**:


everettraven:
@atoato88 this PR should now be in a good state to review. Please let me know if there is anything else needed, thanks!

doc.go
```diff
@@ -18,7 +18,7 @@ limitations under the License.
 // declarative Kubernetes operators to manage the lifecycle of a
 // deployment.
 //
-// Getting Started
+// # Getting Started
```
atoato88, line 21: Is it paragraph markup on markdown format?

pkg/patterns/addon/doc.go
```diff
@@ -26,7 +26,7 @@ implements the addonsv1alpha1.CommonObject interface. The object represents the
 to deploy an instance of a specific Addon in the cluster. This pattern manages a
 Kubernetes deployment for the specific addon based on the Addon Object.
 
-Writing an Addon Operator
+# Writing an Addon Operator
```
atoato88, line 29: same comment above.

atoato88:
Thank you for updating!
It looks good!
And CI tests have passed too.

atoato88 (commented):
I left some comments about `doc.go`. 
I think it's something format update for documentation, but please let me know a background about this update.

doc.go
everettraven: This has to do with `gofmt` updates related to bumping the Go version to Go 1.19.

If you take a look at the **Headings** section of this documentation https://tip.golang.org/doc/comment you can see that as of Go 1.19 the new heading format is similar to Markdown and uses the `#` prefix:

> The # syntax was added in Go 1.19. Before Go 1.19, headings were identified implicitly by single-line paragraphs satisfying certain conditions, most notably the lack of any terminating punctuation.

> Gofmt reformats [lines treated as implicit headings](https://github.com/golang/proposal/blob/master/design/51082-godocfmt.md#headings) by earlier versions of Go to use # headings instead. If the reformatting is not appropriate—that is, if the line was not meant to be a heading—the easiest way to make it a paragraph is to introduce terminating punctuation such as a period or colon, or to break it into two lines. 

I hope this helps!

pkg/patterns/addon/doc.go
everettraven: See: https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/238#discussion_r964838580

doc.go
atoato88: Thank you for detailed and helpful comment.
I understood that. 👍

PR #242, opened 2022-10-05
More mock testing

justinsb:
Create some golden tests to try to make sure our apply library is reasonably efficient / so we can track what it is doing.

- tests: Improve mock kubeapiserver, introduce test for direct Applier
- tests: record and validate the requests we are making
- tests: add test for when objects already exist
- tests: add test for the case when last-applied is present

### DIFF
--- a/pkg/mocks/mockkubeapiserver/getresource.go
+++ b/pkg/mocks/mockkubeapiserver/getresource.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mockkubeapiserver
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+)
+
+// resourceRequestBase holds the common field for single-resource requests
+type resourceRequestBase struct {
+	baseRequest
+
+	Group     string
+	Version   string
+	Resource  string
+	Namespace string
+	Name      string
+
+	SubResource string
+}
+
+// getResource is a request to get a single resource
+type getResource struct {
+	resourceRequestBase
+}
+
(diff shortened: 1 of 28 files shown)


justinsb:
Most of the lines of code are from https://github.com/kubernetes/kops/tree/master/pkg/applylib/mocks/mockkubeapiserver ...  I'm trying "copy and paste" reuse until we figure out where it should live.  But maybe it should just live here, so maybe I should give it its own go.mod here...

pkg/mocks/mockkubeapiserver/getresource.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return err
+	}
+	if !found {
+		return req.writeErrorResponse(http.StatusNotFound)
+	}
+
+	j, err := json.Marshal(object)
+	if err != nil {
+		klog.Warningf("object does not marshal: %v", err)
```
cheftako, line 63: Is there a reason we aren't using structured logging?

pkg/mocks/mockkubeapiserver/putresource.go
```diff
@@ -0,0 +1,80 @@
...
+	resourceRequestBase
+}
+
+// Run serves the http request
+func (req *putResource) Run(ctx context.Context, s *MockKubeAPIServer) error {
+	gr := schema.GroupResource{Group: req.Group, Resource: req.Resource}
+
+	id := types.NamespacedName{Namespace: req.Namespace, Name: req.Name}
+
+	existing, found, err := s.storage.GetObject(ctx, gr, id)
```
cheftako, line 42: existing makes me think its a bool. oldVal? existingVal?

pkg/patterns/declarative/pkg/applier/direct.go
```diff
@@ -46,8 +46,14 @@ func (d *directApplier) NewBuilder(opt ApplierOptions) *resource.Builder {
 	return resource.NewBuilder(restClientGetter)
 }
 
-func (d *directApplier) NewFactory() cmdutil.Factory {
-	return cmdutil.NewFactory(&genericclioptions.ConfigFlags{})
+func (d *directApplier) NewFactory(opt ApplierOptions) cmdutil.Factory {
+	var configFlags genericclioptions.ConfigFlags
+	if opt.RESTConfig != nil {
```
cheftako, line 51: Might be worth a comment on why this is important.
I believe this is preventing unnecessary repeated openapi calls when a restconfig exists and has already made those calls.

cheftako:
/lgtm
Couple of minor questions but on the whole looks good.

pkg/mocks/mockkubeapiserver/putresource.go
justinsb: Will change to existingVal - thank you!

pkg/mocks/mockkubeapiserver/getresource.go
justinsb: Blocked on logr not supporting Warning: https://github.com/go-logr/logr/issues/151

PR #247, opened 2022-10-06
Direct applier: Avoid fetching OpenAPI

justinsb:
We don't normally need it, and it is expensive to fetch.  Use ApplyOptions directly rather than going through the kubectl abstraction layer (which always fetches the OpenAPI)

Builds on #246


justinsb:
/test pull-declarative-test

Known flakiness, addressed in #247

PR #248, opened 2022-10-06
Introduce applyset applier

justinsb:
The applyset applier relies on server-side-apply, which means we don't need as many layers and won't accidentally fetch the OpenAPI every time without knowing (whoops!)


johnsonj:
This is an awesome start. I can think of multiple controller/deployment use cases where I would like to use this!

justinsb:
Thanks @johnsonj - as this is additive (shouldn't change existing code paths) and I got tests to pass I removed the WIP designation!

The huge line count is (mostly) because of the go.sum files - I always wonder if I should be checking those in.  I could split the move of mockkubeapiserver to its own module into a separate PR, but it isn't the bulk of the line count.

I have a few more PRs "up next" where I use this for the metrics_test, which requires supporting watch, so mockkubeapiserver is getting better - hopefully good enough for your use case when you get there!

johnsonj:
I think it's harmless to include go.sum. I'm no expert in the ever-evolving world of go modules.. but at least we'll know CI will run exactly the code as expected :) 

/lgtm

PR #249, opened 2022-10-06
tests: switch to controller restmapper

justinsb:
The discovery rest mapper is non-deterministic, which makes it hard to use with our tests.

Builds on #247


johnsonj (commented):
/lgtm

PR #250, opened 2022-10-11
Use mockkubeapiserver in metrics_test

justinsb:
A few changes needed:

* mockkubeapiserver doesn't support patching of maps (yet), as that
  depends on the OpenAPI, so we switch to ConfigMap and
  Server-Side-Apply.

* the metrics test relies on watch, so we add a basic implementation
  of watch to the mockkubeapiserver


johnsonj (commented):
/lgtm

PR #255, opened 2022-10-19
Create nil-safe wrapper around Items

justinsb:
Avoids a potential panic.


johnsonj:
@justinsb There's a few other usages of Items directly, consider making the value private and moving all callers over to `GetItems`

- https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/beb06a5d4ddd3387c1aaa7694a587462da5c58e6/pkg/patterns/declarative/pkg/applier/applylib.go
- https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/4ba4cadcd4ca182c7b29c0569c85fbbd2996c762/pkg/patterns/declarative/application.go


/lgtm

PR #258, opened 2022-10-19
Log error when we can't get the GVK for an object

justinsb:
Should not be a behavioural change, but we can log a better message.


johnsonj (commented):
/lgtm

PR #259, opened 2022-10-19
mockkubeapiserver: don't update if the object doesn't change

justinsb:
This is what the real apiserver does, and it's important because it
means we don't send another watch notification / don't update the
resourceVersion.


johnsonj (commented):
/lgtm

PR #260, opened 2022-10-19
Create test reconciler and capture http requests

justinsb:
This enables us to test full controllers very easily, including the exact requests we make.

~WIP while I figure out the last bits of non-determinism here.~


justinsb:
This no longer seems to be non-deterministic - turning off the metrics (which starts watchers) was the last thing here :tada: 

/assign @barney-s

pkg/test/testreconciler/simpletest/controller_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+	mgrContext, mgrStop := context.WithCancel(ctx)
+	go func() {
+		time.Sleep(2 * time.Second)
+		// time.Sleep(200 * time.Second)
+		mgrStop()
+	}()
+	if err := mgr.Start(mgrContext); err != nil {
+		h.Fatalf("error starting manager: %v", err)
+	}
```
barney-s, line 140: Do we have to wait for one reconcile to complete and http logging to complete ?

barney-s (approved):

pkg/test/testreconciler/simpletest/controller_test.go
justinsb: So it would be nice to reach a steady state, right now I'm just hard-coding a 2 second delay.  We do want to know if it enters a busy loop, so we don't want to stop after the first reconcile.  Ideally we would stop when the system became idle - no watch events, no pending reconciles; I can look at getting there but that'll be a separate PR I think!

PR #264, opened 2022-11-07
Allow applier to be overridden per-controller

justinsb:
This is useful in tests and probably in real-world also.


johnsonj (commented):
Seems like the right place to get it!

/lgtm

atoato88:
LGTM

PR #265, opened 2022-11-07
testing: record http responses also

justinsb:
- testing: record http responses also
- testdata: update to include responses also


johnsonj (commented):
/lgtm

PR #266, opened 2022-11-07
Report status as unhealthy if we failed to reconcile

justinsb:
A little misleading to do otherwise!

### DIFF
--- a/pkg/patterns/addon/pkg/status/aggregate.go
+++ b/pkg/patterns/addon/pkg/status/aggregate.go
@@ -45,11 +45,15 @@ type aggregator struct {
 	client client.Client
 }
 
-func (a *aggregator) Reconciled(ctx context.Context, src declarative.DeclarativeObject, objs *manifest.Objects, _ error) error {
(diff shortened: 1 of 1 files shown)


pkg/patterns/addon/pkg/status/aggregate.go
```diff
@@ -45,12 +45,16 @@ type aggregator struct {
...
 
-func (a *aggregator) Reconciled(ctx context.Context, src declarative.DeclarativeObject, objs *manifest.Objects, _ error) error {
+func (a *aggregator) Reconciled(ctx context.Context, src declarative.DeclarativeObject, objs *manifest.Objects, reconcileErr error) error {
 	log := log.FromContext(ctx)
 
 	statusHealthy := true
 	statusErrors := []string{}
 
+	if reconcileErr != nil {
+		statusHealthy = false
```
johnsonj, line 55: Do you also want to propagate the message to the user? It's a little blurry here as its sort of internal, but it would aid in debugging.

If so add:
statusErrors = append(statusErrors, fmt.Sprintf("%v", reconcileErr))

johnsonj (commented):
/lgtm

PR #267, opened 2022-11-07
mockkubeapiserver: fix bug where comparing wrong objects

justinsb:
We were trying to avoid spurious updates, but we ended up ignoring all
updates (?).


johnsonj (commented):
/lgtm

PR #268, opened 2022-11-07
applylib: Recognize and map the --force kubectl option

justinsb:
We use the force option quite a lot, so we need to support it.

Also nudge away from ExtraArgs towards explicit options.


johnsonj (commented):
/lgtm

atoato88:
LGTM

PR #269, opened 2022-11-07
Record event when reconciler fails unexpectedly

justinsb:
Otherwise it's difficult to track down what has gone wrong.

This may lead to over-eventing (if we've already recorded the error),
but we can deal with that problem when we come to it!


johnsonj:
Awesome!
/lgtm

atoato88:
LGTM

PR #272, opened 2022-11-09
Better status functions

justinsb:
We pass in a function to query the current state of objects, this
allows us to cache the results (if using SSA) and allows for cluster
targeting.

### DIFF
--- a/pkg/patterns/declarative/pkg/manifest/objects.go
+++ b/pkg/patterns/declarative/pkg/manifest/objects.go
@@ -27,6 +27,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
@@ -318,6 +319,16 @@ func (o *Object) GroupKind() schema.GroupKind {
 
 func (o *Object) GroupVersionKind() schema.GroupVersionKind {
 	return o.object.GroupVersionKind()
+}
+
+// NamespacedName returns the name and namespace of the object in a types.NamespacedName.
+// Note that this reflects the state of the object; if the namespace is not yet set,
+// it will returned as "" here, even though it would likely be defaulted before apply.
+func (o *Object) NamespacedName() types.NamespacedName {
+	return types.NamespacedName{
+		Namespace: o.GetNamespace(),
+		Name:      o.GetName(),
+	}
 }
 
 func (o *Objects) JSONManifest() (string, error) {
--- a/pkg/patterns/declarative/reconciler.go
+++ b/pkg/patterns/declarative/reconciler.go
@@ -21,16 +21,15 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
+	"reflect"
 	"strings"
 
-	"k8s.io/apimachinery/pkg/api/meta"
-	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative/kustomize"
-
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
@@ -42,6 +41,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/kustomize/kyaml/filesys"
 
+	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/addon/pkg/utils"
+	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative/kustomize"
 	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative/pkg/applier"
 	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative/pkg/manifest"
 )
@@ -116,7 +117,7 @@ func (r *Reconciler) Init(mgr manager.Manager, prototype DeclarativeObject, opts
 
 	r.restMapper = mgr.GetRESTMapper()
 
-	if err = r.applyOptions(opts...); err != nil {
+	if err := r.applyOptions(opts...); err != nil {
 		return err
 	}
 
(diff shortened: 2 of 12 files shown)


pkg/patterns/declarative/reconciler.go
```diff
@@ -170,26 +175,60 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
...
+			log.Error(err, "error getting status")
+			return result, err
+		}
+		newStatus, err := utils.GetCommonStatus(instance)
+		if err != nil {
+			log.Error(err, "error getting status")
+			return result, err
+		}
+		if !reflect.DeepEqual(oldStatus, newStatus) {
+			if err := r.client.Status().Update(ctx, instance); err != nil {
```
JeffLuoo, line 214: Q: How would the `Status().Update(..)` resolves update conflict when the user is modifying the value in the CR?

justinsb: So the user doesn't normally modify the status.  But in any case, the call to Update actually uses optimistic concurrency.  We provide the same `instance` that we read.  That instance contains a ResourceVersion.  If there was a concurrent change, the resourceVersion on the server does not match, and the Update is rejected.  That is returned as an error, we return an error from the reconciler, and the reconciler is run again (with backoff).

It doesn't normally happen because this controller _should_ (generally) own the Status.  So we don't handle it specially.

JeffLuoo: Thank you for the explanation.

atoato88:
/lgtm
Sorry for my delay reviewing.

atoato88:
Oh, this PR needs rebase..

@justinsb 
Could you rebase?

pkg/patterns/declarative/pkg/applier/type.go
```diff
@@ -5,14 +5,15 @@ import (
...
+	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative/pkg/manifest"
 )
 
 type Applier interface {
 	Apply(ctx context.Context, options ApplierOptions) error
 }
 
 type ApplierOptions struct {
-	Manifest string
+	Objects []*manifest.Object
```
yuwenma, line 16: Do you mind explaining a little bit why this replaces the Manifests with `manifest.Object` slice?  I look at all the places where use `ApplierOption.Objects`, it seems to have no logic change besides the two types conversion. 

From a style perspective, I like the change (typed resource group vs serialized value), just to ask to double check in case I miss any feature changes.

pkg/patterns/declarative/pkg/manifest/objects.go
```diff
@@ -305,6 +306,13 @@ func (o *Object) GroupVersionKind() schema.GroupVersionKind {
 	return o.object.GroupVersionKind()
 }
 
+func (o *Object) NamespacedName() types.NamespacedName {
+	return types.NamespacedName{
+		Namespace: o.GetNamespace(),
```
yuwenma, line 311: Reading the `types.NamespacedName` doc, it says the `Namespace` is "[mandatory](https://github.com/kubernetes/apimachinery/blob/1cc21f9c83e1ffc69aa85af3bc0cc7467e9e5113/pkg/types/namespacedname.go#L19)", will this be applied to this function? The name "NamespacedName" seems to tell the specific `manifest.Object` must be a namespace scoped resource. 

Another follow up question about the ambiguity of the namespace value. I see this method in this PR is only used in `StatusBuilder`, which namespace value  is very clear (the value can be "default" or "SOME_CUSTOM_NS" for namespace socped and can be empty for cluster scoped resource). But because `manifest.Object` is Public,  users can call it before the manifest is applied, then the empty `namespace` value could be ambiguous between un-assigned namespace scope resource and cluster scoped resource (which both are empty string). This seems to be a risk condition, if it is valid we can at least call it out in the function docstring.

yuwenma:
/LGTM 

Two questions and the PR looks great. I'm very looking forward to use this feature in my operator!

pkg/patterns/declarative/pkg/applier/type.go
justinsb: That was actually https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/275 , it was included here because github doesn't have chained PRs.  Sorry - I should have said "builds on #275" or similar.

The actual problem was that I was passing yaml to something that expected json.  They are both `string` (or `[]byte`).  To prevent that confusion in future, I switched to typed objects.

yuwenma: That's fair! Thanks for the explanation.

pkg/patterns/declarative/reconciler.go
```diff
@@ -198,42 +236,41 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 	objects, err := r.BuildDeploymentObjectsWithFs(ctx, name, instance, fs)
 	if err != nil {
 		log.Error(err, "building deployment objects")
-		return nil, fmt.Errorf("error building deployment objects: %v", err)
+		return statusInfo, fmt.Errorf("error building deployment objects: %v", err)
 	}
+	statusInfo.Manifest = objects
```
yuwenma, line 241: Super nit:
The logic is a little bit confusing considerin that [ln222/259](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/272/files#diff-91d9000f2817ed2698caf7b294e34ed1bc739bf60b7dc1af3136055b18741312R259) `objects, err = parseListKind(objects)` and then 
[ ln207/264](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/272/files#diff-91d9000f2817ed2698caf7b294e34ed1bc739bf60b7dc1af3136055b18741312R264) has duplicate `statusInfo.Manifest = objects`

Maybe moving the logic[ ln259-264](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/272/files#diff-91d9000f2817ed2698caf7b294e34ed1bc739bf60b7dc1af3136055b18741312R259-R264) here. Otherwise it looks like a duplication of  ln 264 and not obvious enough why the code appends `Objects` to Manfiests multiple times.

pkg/patterns/declarative/reconciler.go
```diff
@@ -656,6 +719,9 @@ func (r *Reconciler) CollectMetrics() bool {
 	return r.options.metrics
 }
 
+// GetObjectFromCluster gets the current state of the object from the cluster.
+//
+// deprecated: use LiveObjectsFunc instead when computing status
```
yuwenma, line 724: Do you mean the `LiveObjectReader` in this PR or in some code in other feature branch? I can't find a `LiveObjectFunc` else where [in the master k-d-p branch ](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/search?q=LiveObjectsFunc)

yuwenma (approved):

justinsb:
/retest

Going to roll the dice one more time, because I think this is the uncertain ordering between watch & apply, so we're closer to fixing it if this goes in, but if that is true I would expect the tests to pass about half the time.

yuwenma:
For watch/update conflict, would it help if we bump the test manager's cancelling period https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/272/files#diff-62091d2b32c6b6c94aaed7ea3055732f56d47b34c5f66f31079a61a3908c3689R141 , assuming the probability of happening a second status update (after watch) increases?

pkg/patterns/declarative/pkg/manifest/objects.go
justinsb: Good points.  Added a docstring.  If we can find a better struct or if you have a better name, I'd be very open to one!

We aren't actually using this in many places (yet, though I think the plan was to use it in more places) so we could also just remove it.  I propose we leave it for now and see if it is helpful.

pkg/patterns/declarative/reconciler.go
justinsb: You're right - this logic is super confusing.  I think it has "evolved" over time and some entropy has snuck in, and we should clean it up.

I will send a separate PR if that's OK - your point is very valid but I think it's a separate concern to the logic in this PR (i.e. this problem already existed)

pkg/patterns/declarative/reconciler.go
justinsb: Good point - I think I renamed it a few times - I do mean LiveObjectReader.  Thank you!

PR #274, opened 2022-11-15
Prune whitelist

stephanhorsthemke:
This PR fixes some odds and ends to make Pruning + Allowlisting CRDs for pruning work. We need it to use pruning in our custom controller.

This actually fixes the `--prune-whitelist` part of #201 

I put some comments in the diff. I believe it would be nice for us already to make CascadingStrategy configurable. I hardcoded it to `Foreground` here but we don't know yet what behaviour we wanna have.

**Additional documentation**:


pkg/patterns/declarative/pkg/applier/direct.go
```diff
@@ -128,16 +128,19 @@ func (d *DirectApplier) Apply(ctx context.Context, opt ApplierOptions) error {
 	for i, arg := range opt.ExtraArgs {
 		switch arg {
 		case "--force":
+			// TODO Does this do anything? It seems like opt (aka ApplierOptions) is not used anymore
```
justinsb, line 131: Good question.  I _think_ it's right?  Here in this loop we map from ExtraArgs (which are very oriented towards the command line execution which we're trying to stop using) to explicit flags like `Force`.  Then below we map `applyOpts.ForceConflicts = opt.Force`.  And then we pass applyOpts to `Run`.  But I might be missing something?

We do have some fancy tests that can record the HTTP calls, so we could actually test the resulting behaviour now, but I _think_ it's fine?

pkg/patterns/declarative/pkg/applier/direct.go
```diff
@@ -128,16 +128,19 @@ func (d *DirectApplier) Apply(ctx context.Context, opt ApplierOptions) error {
...
 			opt.Force = true
 		case "--prune":
 			applyOpts.Prune = true
 		case "--selector":
 			applyOpts.Selector = opt.ExtraArgs[i+1]
+		case "--prune-whitelist":
+			applyOpts.PruneWhitelist = append(applyOpts.PruneWhitelist, opt.ExtraArgs[i+1])
```
justinsb, line 138: Suggestion (non-blocking): You might consider exposing a strongly-typed form like `PruneWhitelist []schema.GroupVersionKind` or `[]schema.GroupKind`, more like what we're doing with `Force`

pkg/patterns/declarative/pkg/applier/direct_test.go
```diff
@@ -152,26 +152,34 @@ metadata:
...
+  name: foo-operator
+  namespace: kube-system`,
+					args: []string{"--prune", "--selector", "label=true", "--prune-whitelist", "core/v1/Namespace"},
+					// TODO is this used somewhere?
+					expectApplyOptions: &apply.ApplyOptions{
+						Prune: true,
+						PruneWhitelist: []string{"core/v1/Namespace"},
+					},
+					expectCheckFunc: func(opt *apply.ApplyOptions) error {
+						if opt.Prune != true {
```
justinsb, line 171: Suggestion (non-blocking): You might try using `cmp.Diff` here; it's pretty magical ... when it works!  It's sort of like `reflect.DeepEquals` but it gives you a difference you can print in tests.  Saves you the work of doing the manual comparison.  I say "when it works" because it can fall down with complex types and require additional arguments, but this type looks simple enough that I think it will "just work".

pkg/patterns/declarative/reconciler.go
```diff
@@ -302,7 +302,7 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 		}
 	}
 
-	applyOpt := applier.ApplierOptions{
+	applierOpt := applier.ApplierOptions{
```
justinsb, line 305: Good call, thanks for fixing!

justinsb (commented):
Looks good if you want to remove Draft!  Some suggestions if you want to do more in this PR, but they can also be done separately or not at all :-)

stephanhorsthemke:
Thanks for the review already @justinsb !
I found that it was not working. So the applier test did but when I used it in a controller still nothing happened. Did not run prune at all.
 I believe I found the two problems which I just fixed in the last commit. Will try again running it now and then look at whether I can tackle your improvement :)

pkg/patterns/declarative/pkg/applier/direct.go
stephanhorsthemke: Totally correct. I missed the line where we map `opt.Force` to  `applyOpts.ForceConflicts`. So this makes sense. One could just refactor and set `applyOpts.ForceConflicts` to `true` in this line I guess.

pkg/patterns/declarative/pkg/applier/direct_test.go
stephanhorsthemke: Cool suggestion! Haven't used it yet. I gave it a shot but it seemed like it is not the most straighforward case as `prune.PruneResource` has unexported fields. The panic message suggested to create a complex type. This would definitely be cool.

pkg/patterns/declarative/pkg/applier/direct.go
```diff
@@ -76,13 +77,14 @@ func (d *DirectApplier) Apply(ctx context.Context, opt ApplierOptions) error {
 	b := d.inner.NewBuilder(opt)
 	f := d.inner.NewFactory(opt)
 
+	dynamicClient, err := f.DynamicClient()
+	if err != nil {
+		return err
+	}
+
 	if opt.Validate {
 		// This potentially causes redundant work, but validation isn't the common path
```
stephanhorsthemke, line 86: I wonder if this redundant work is comment is related to the creation of the DynamicClient. If so this comment should be deleted I guess. Also, is it fine to create that client every time this runs or should we be smarter about it?

pkg/patterns/declarative/reconciler.go
```diff
@@ -302,18 +302,20 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
...
+		RESTConfig: 		r.config,
+		RESTMapper: 		r.restMapper,
+		Namespace:  		ns,
+		Manifest:   		manifestStr,
+		Validate:   		r.options.validate,
+		ExtraArgs:  		extraArgs,
+		Force:      		true,
+		// TODO Make this configurable
+		CascadingStrategy: 	"Foreground",
```
stephanhorsthemke, line 314: I'd really like to have this configurable. ATM we don't even know what we wanna have here so we might not be happy with "Foreground" going forward.
I first thought I could create a new interface like `PruneWhiteLister`, but this is maybe something for another PR? And then I wonder if those Interfaces are the way to go?
Would it be possible to have a more general approach to allow to configuration of more of those fields and would that make sense?

pkg/patterns/declarative/pkg/applier/direct.go
stephanhorsthemke: I needed to use `applyOpts.PruneResources` anyways. `PruneWhitelist` did not seem to have the desired (or any) effect.

pkg/patterns/declarative/pkg/applier/direct.go
justinsb: I think I wrote the original comment.  And yes, I think we're constructing new clients and probably doing discovery again, but in general I thought it wasn't as critical to optimize given that validation isn't as useful as it once was - the server should be doing the same validation here, we don't expect to fail validation given that we're probably constraining the manifests etc.

Maybe we should update the comment to say something like "validation likely makes redundant apiserver requests and is less optimized than the non-validation case, but validation isn't the common path" ?

pkg/patterns/declarative/reconciler.go
justinsb: Putting it into the ApplierOptions is a good start :-)  We should think about mapping this field in the kubectl applier though, as in theory people are still using that (and once the applyset applier is more ready we would do that, but right now I don't think that supports prune anyway)

We could have something like `r.options.validate`, but we could also think about creating a set of the configurable apply options.  I have a PR out for "hooks" that I think would allow this to be changed through code, but as we control the implementation here we can choose what user-experience (developer-experience) we want to create.

I think what we would ideally do is wait till the objects are actually deleted - I don't know if _any_ of the cascadingstrategies actually do that (?), but kubectl can wait so it should be possible...

pkg/patterns/declarative/pkg/applier/direct_test.go
justinsb: Thanks for trying!  And yes, the unexported fields are always the problem :-(  There are some options to control this but it's a lot more complex - sorry, I thought they were "simple" all-public structs.

pkg/patterns/declarative/pkg/applier/type.go
```diff
@@ -19,6 +20,11 @@ type ApplierOptions struct {
 	Namespace  string
 	Validate   bool
 
+	CascadingStrategy metav1.DeletionPropagation
+
+	PruneWhitelist []string
```
justinsb, line 25: Nit: We probably should change this to something like `PruneResources` to avoid non-inclusive language (because this becomes part of our API I think, so renaming it later is harder)

justinsb (approved):

justinsb:
A few thoughts but LGTM - thanks for the contribution @stephanhorsthemke 

/approve
/lgtm

justinsb:
Looks like you need to run goimports (or gofmt) on a few files:

```
Error: goimports should be run on these files:
pkg/patterns/declarative/pkg/applier/direct.go
pkg/patterns/declarative/pkg/applier/direct_test.go
pkg/patterns/declarative/pkg/applier/type.go
pkg/patterns/declarative/reconciler.go
```

pkg/patterns/declarative/pkg/applier/direct.go
stephanhorsthemke: Understood! Sounds good to me!

pkg/patterns/declarative/reconciler.go
stephanhorsthemke: About the cascading strategy, reading [the docs](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#cascading-deletion).
`Background` is the default usually (however there was no default set when I ran it, it resulted in an error complaining that the propagationPolicy which ultimately is the cascading strategy can't be `/`). 
I do believe that we in our controller want to wait and agree with you that this makes probably the most sense to be the default. And I believe that it is what the `Foreground` strategy does (from docs: `The object remains visible through the Kubernetes API until the deletion process is complete.`)
I haven't been able to really see the difference when trying it out yet.

stephanhorsthemke: I understand that we probably also need to make some changes in the other applies, `exec.go`!? I don't really know if you want like to see this changes in this PR already. Will have a look at least :)

stephanhorsthemke: Looking at that applier I see multiple reasons why pruning won't work. Some of which I fixed for the direct applier, some more. On the one hand I feel that I know a decent amount about it now to probably fix all of it quite swiftly, but does it make sense?  I could take half an hour or so to make a draft PR building on top of this to do everything I know to at least have a quick start?

stephanhorsthemke: I have no idea how the `manifest with prune` test in `exec_test.go` can even work :D
For sure the `expectedArgs` will not result in the expected behavior (pruning). As you can't use `-f` and `--prune`.

That was one of my confusions in the direct applier. If you set force in the direct applier it is not gonna result in the `-f` behaviour but it results in setting `ForceConflicts` aka `--force-conflict`. Thinking about this I believe this is quite wrong

pkg/patterns/declarative/pkg/applier/type.go
stephanhorsthemke: Totally agree! I have thought about this.
But, 
1. if we call it `PruneResources` we have a different type for the same field in `ApplierOptions` than in `ApplyOpts` which probably is confusing.
2. We could call it `PruneAllowList`, I have seen this somewhere. But then it is confusing as we go from the interface `PruneWhitelister` to `PruneAllowList` to `PruneResources`. 

This detail is so hard.
Maybe this is actually the time to change `PruneWhitelister` to `PruneAllowLister`? I believe it has never worked anyways up until now? Probably that's still not a cool thing to do though.

justinsb: Naming is always the hardest thing!  Let's do it as its own PR, so we can think about it in isolation (including the consequences in terms of breaking changes!)

justinsb:
/ok-to-test

(I'm assuming that's why the prow tests aren't running?)

justinsb:
Passing tests - thanks for this @stephanhorsthemke 

/approve
/lgtm

PR #276, opened 2022-11-16
httprecorder: recognize and stream watch responses

justinsb:
Otherwise we were trying to buffer them entirely, which wasn't working well.

Note that we now also capture the second reconcile, triggered by the status update.


yuwenma (approved):
Just curious: is the test data auto-generated somehow (in pkg/test/testreconciler/simpletest/testdata/reconcile/direct/create/expected-http.yaml)?

PR #277, opened 2022-11-16
Add support for hooks to the reconciler

justinsb:
Hooks are an extension of the sink concept that we already use for
watches, but support more interception points and we can have multiple
hooks.


atoato88 (commented):
Almost LGTM, just one thing.
Please check the comment.

atoato88:
@justinsb 
Is this PR still live?
This PR needs rebase, could you do that?

yuwenma (approved):

yuwenma:
/LGTM

Everything looks good except the CI test failure.

justinsb:
> Is this PR still live?

Yes, sorry, I've just been stuck trying to figure out how to get all these tests all working again.  The testing became so strict that it because impossible to merge code.

That said, I think I've got it now... we need to normalize APIResources (#312), we need to normalize discovery requests because kubectl doesn't do them in any predictable order (#313), and then we need to update everything (#307), and then finally I will change this PR to set up the watches in BeforeApply instead of AfterApply (which is actually really easy with this PR).

And then I think we're back in business :-)

PR #278, opened 2022-11-16
mockkubeapiserver: register built-in types

justinsb:
We generate a metadata file that includes the additional type
information we need.


mockkubeapiserver/kubernetes_builtin_schema.go
```diff
@@ -5,35 +5,79 @@ import (
...
+	defer c.mutex.Unlock()
 
-	if cacheKubernetesBuiltinSchema != nil {
-		return cacheKubernetesBuiltinSchema, nil
+	if schema := c.schemas[key]; schema != nil {
+		return schema, nil
 	}
 
-	schemaYAML := kubernetesBuiltinSchemaYAML
+	if key != BuiltinKey {
```
apelisse, line 58: I don't get it, if the key is NOT "BuiltInKey", then this will fail, but then how do you get anything but "BuiltinKey" in the list?

justinsb: That's the only one that works at the moment.  But I thought this was clearer (externally) and can allow for more in future...  It's not great, but maybe it's better than the multiple globals we had before.

PR #280, opened 2022-11-19
Add AddAnnotations companion to AddLabels

tomasaschan:
I had a use case for setting annotations on all objects in an object transform, and found that while we do have `AddLabels` that does this, there's no corresponding `AddAnnotations`.

There _is_ a way to do this without this PR, since the `manifest.Object` exposes `UnstructuredObject()` to get at the underlying `unstructured.Unstructured`, but its docstring states

> UnstructuredContent exposes the raw object, primarily for testing

so I figured it's better to not ask users to take a dependency on that.

### DIFF
--- a/pkg/patterns/declarative/annotations.go
+++ b/pkg/patterns/declarative/annotations.go
@@ -11,7 +11,6 @@ import (
 func AddAnnotations(annotations map[string]string) ObjectTransform {
 	return func(ctx context.Context, o DeclarativeObject, manifest *manifest.Objects) error {
 		log := log.Log
-		// TODO: Add to selectors and labels in templates?
 		for _, o := range manifest.Items {
 			log.WithValues("object", o).WithValues("annotations", annotations).V(1).Info("add annotations to object")
 			o.AddAnnotations(annotations)
--- a/pkg/patterns/declarative/annotations.go
+++ b/pkg/patterns/declarative/annotations.go
@@ -1,0 +1,22 @@
+package declarative
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative/pkg/manifest"
+)
+
+// AddAnnotations returns an ObjectTransform that adds annotations to all the objects
+func AddAnnotations(annotations map[string]string) ObjectTransform {
+	return func(ctx context.Context, o DeclarativeObject, manifest *manifest.Objects) error {
+		log := log.Log
+		// TODO: Add to selectors and labels in templates?
+		for _, o := range manifest.Items {
+			log.WithValues("object", o).WithValues("annotations", annotations).V(1).Info("add annotations to object")
+			o.AddAnnotations(annotations)
+		}
+
+		return nil
+	}
+}
--- a/pkg/patterns/declarative/pkg/manifest/objects_test.go
+++ b/pkg/patterns/declarative/pkg/manifest/objects_test.go
@@ -281,6 +281,69 @@ spec:
 	}
 }
 
+func Test_AddAnnotations(t *testing.T) {
+	inputManifest := `---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: test-app
+  name: frontend
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: test-app
+  strategy: {}
+  template:
+    metadata:
+      labels:
+        app: test-app
+    spec:
+      containers:
+      - image: busybox
+        name: busybox`
+	tests := []struct {
+		name                string
+		inputManifest       string
+		inputAnnotations    map[string]string
+		expectedAnnotations map[string]string
+	}{
+		{
+			name:                "add annotations which are all new one",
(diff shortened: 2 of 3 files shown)


tomasaschan:
For full context, here's the simplification this PR would enable.

Without this PR, I have to do this:
```
	return func(ctx context.Context, do declarative.DeclarativeObject, os *manifest.Objects) error {
		errs := &multierror.Error{}

		for _, o := range os.Items {
			merged := make(map[string]string)
			for k, v := range o.UnstructuredObject().GetAnnotations() {
				merged[k] = v
			}

			for k, v := range annotationMaker(ctx, do) {
				merged[k] = v
			}

			for k, v := range merged {
				if err := o.SetNestedField(v, "metadata", "annotations", k); err != nil {
					errs = multierror.Append(errs, err)
				}
			}
		}
		return errs.ErrorOrNil()
	}
```

With this PR, the following is equivalent:
```
	return func(ctx context.Context, do declarative.DeclarativeObject, os *manifest.Objects) error {
		for _, o := range os.Items {
			annotations := annotationMaker(ctx, do)
			o.AddAnnotations(annotations)
		}
		return nil
	}
```
Note that not only is the code simpler - by avoiding to use `SetNestedFields`, it's also safer!

pkg/patterns/declarative/annotations.go
```diff
@@ -0,0 +1,22 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative/pkg/manifest"
+)
+
+// AddAnnotations returns an ObjectTransform that adds annotations to all the objects
+func AddAnnotations(annotations map[string]string) ObjectTransform {
+	return func(ctx context.Context, o DeclarativeObject, manifest *manifest.Objects) error {
+		log := log.Log
+		// TODO: Add to selectors and labels in templates?
```
justinsb, line 14: I guess this was copy-pasted from AddLabels?  I don't think it applies to annotations.

justinsb (commented):
LGTM (with one small nit about an extra TODO - not sure if you want to fix).

And yes, we really shouldn't be exposing UnstructuredObject, because it doesn't clear the cached json.  Maybe we should just clear the cached json in UnstructuredObject, to be safe in case someone tries using it outside a test, because otherwise I think the behaviour could be really confusing!

pkg/patterns/declarative/pkg/manifest/objects_test.go
```diff
@@ -281,6 +281,69 @@ spec:
 	}
 }
 
+func Test_AddAnnotations(t *testing.T) {
+	inputManifest := `---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
```
justinsb, line 289:
```suggestion
  annotations:
```

pkg/patterns/declarative/pkg/manifest/objects_test.go
```diff
@@ -281,6 +281,69 @@ spec:
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := context.Background()
+			objects, err := ParseObjects(ctx, tt.inputManifest)
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			for _, o := range objects.Items {
+				o.AddAnnotations(tt.inputAnnotations)
+				if len(tt.expectedAnnotations) != len(o.object.GetLabels()) {
```
justinsb, line 336:
```suggestion
				if len(tt.expectedAnnotations) != len(o.object.GetAnnotations()) {
```

justinsb (commented):
Some suggested changes to get the tests to pass

pkg/patterns/declarative/annotations.go
tomasaschan: Yeah, it was a copy-paste.
```suggestion
```

tomasaschan:
> And yes, we really shouldn't be exposing UnstructuredObject, because it doesn't clear the cached json. 

Yeah, that's why I used `SetNestedField` instead of just modifying the returned `Unstructured` with e.g. `AddAnnotations()`. Maybe it could be nice to instead offer some general function that can be used to modify the `Unstructured`, but with proper cache behavior?
```
type Mutator func(*unstructured.Unstructured) error
func (o *objects.Object) Mutate(f Mutator) error {
  if err := f(o.object); err != nil {
    return err
  }
  o.json = nil
  return nil
}
```
(which has the implicit contract that if an error is returned from the mutator, the object is not modified, but maybe that's OK...)

With something like that, anyone could implement arbitrary transforms like this without worrying about the json cache.

justinsb:
> Maybe it could be nice to instead offer some general function...

Yes, I think that's an excellent idea.  We could also just drop the cache, it's a relatively minor optimization.  If we keep the cache, we should at least comment that mutating with the UnstructuredObject accessor is unsafe.

A nice thing about the mutator with the callback is that we could also support it on type `Objects` (it would presumably visit each object in Items)

justinsb:
Thanks for the contribution @tomasaschan

/approve
/lgtm

tomasaschan:
Thanks for taking the time to review despite me submitting this on a Saturday!

PR #281, opened 2022-11-19
Set Overwrite in ApplyOptions

justinsb:
We lost the default value in #247, which is a regression that prevents
some changes from going through.


tomasaschan (approved):
Not sure my approval counts for anything here, but this is breaking stuff for us and we'd love to get it in ASAP! :)

/lgtm /approved

justinsb:
@tomasaschan I think it _should_ count if you put `/lgtm` on its own line :-)  It should work for any kubernetes org member.  If you're not yet a kubernetes org member, we should do something about that :-)

PR #283, opened 2022-11-30
add option to configure cascading strategy

stephanhorsthemke:
This PR makes the CascadingStrategy added in https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/274 configurable, adding a new option `withCascadingStrategy` and making sure that `Foreground` is the default if not set.

We need it to configure different CascadingStrategies for main and test

Fixes #

I have not seen proper tests I could test this with. Are there some?
I ran this with the controller I am working on.

**Additional documentation**:

### DIFF
--- a/pkg/patterns/declarative/options.go
+++ b/pkg/patterns/declarative/options.go
@@ -19,6 +19,8 @@ package declarative
 import (
 	"context"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative/pkg/applier"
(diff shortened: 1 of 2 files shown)


justinsb:
Sorry for the delay here!  Was trying to get some test coverage but we have a bit of an LGTM backlog on those PRs :-)

I dug into what this does, this sets the argument `propagationPolicy` on delete requests.  So we should be able to pick that up in our tests if we trigger a delete, but we'll just be testing that it is passed through.

One nit is I'm wondering if we should call it `withDeletionPropagationPolicy`... "deletion" to make it clearer that it is only on delete, "propagationPolicy" to mirror the actual flag.  But it's actually easy enough to change later (with an alias for compatibility), and I'm thinking kubectl does call it cascading and not propagationPolicy, and that might be more familiar to users.

So...

/approve
/lgtm

justinsb:
I think we've got tests passing again!

/ok-to-test

I'll also figure out what permissions you need to be able to trigger tests yourselves / have them run automatically!

PR #284, opened 2022-11-30
mockkubeapiserver: support get of cluster-scoped object

justinsb:
An endpoint I had previously missed!


justinsb:
/assign @apelisse 

Joining the "embarassing bugs" club, though this is more "building as we need it" .... maybe.  That's my story, anyway.

PR #287, opened 2022-12-18
tests: normalize http request order

justinsb:
To avoid spurious failures from concurrent requests, we sort
consecutive API discovery requests alphabetically.


johnsonj (commented):
/lgtm

PR #288, opened 2022-12-19
tests: update expected test output for the latest kubectl version

justinsb:
We get the kubectl version from our runner pod, and it has been
updated to a newer version that sends a new Accept header.

Update the golden output to expect that.

We probably should run with a known version of kubectl (or skip the
test if the version is not correct), but first step is at least to
match what we have.


justinsb:
Actually looks like we don't have to tackle this just yet, it's just the version on my machine that has updated.  So need to update the kubectl version used to generate the golden output.

/close

PR #289, opened 2022-12-19
tests: run with a known version of kubectl

justinsb:
Update hack/ci/test.sh to download kubectl and add it to the path.

Add dev/update-golden to generate golden output with the kubectl version.

We run with kubectl 1.26 so that we get the fast-discovery, as this is
expected to roll out more generally over the next few months.

### DIFF
--- a/pkg/patterns/declarative/pkg/applier/exec_test.go
+++ b/pkg/patterns/declarative/pkg/applier/exec_test.go
@@ -144,6 +144,18 @@ metadata:
 }
 
 func TestKubectlApplier(t *testing.T) {
+	kubectlPath, err := exec.LookPath("kubectl")
+	if err != nil {
+		t.Fatalf("failed to find kubectl on path: %v", err)
+	}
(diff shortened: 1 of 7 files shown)


justinsb:
Getting this to pass reliably probably depends on #287

pkg/patterns/declarative/pkg/applier/exec_test.go
```diff
@@ -144,6 +144,18 @@ metadata:
 }
 
 func TestKubectlApplier(t *testing.T) {
+	kubectlPath, err := exec.LookPath("kubectl")
```
barney-s, line 147: would this prefer ./bin over standard paths ?

justinsb: Only if `./bin` is set in the PATH, which we do here :-)

The goal here is primarily to give a nice error if kubectl isn't on the path, but also to print the path to kubectl and the version we found, so that if something weird happens that looks like a kubectl version issue we can figure out what the problem is from the logs!

PR #290, opened 2023-01-04
Update README about required go version

CatherineF-dev:

```
2023/01/04 15:52:14 failed to initialize project: unable to run pre-scaffold tasks of "base.go.kubebuilder.io/v3": go version 'go1.17' is incompatible because 'requires 1.13 <= version < 1.17'. You can skip this check using the --skip-go-version-check flag
```

Fixes #

**Additional documentation**:


justinsb:
Thank you - ideally we would be compatible with later versions of go, but maybe #291 will fix that.  I agree with documenting it in the meantime.

Test failure looks unrelated (and I think we have an open PR for it as well, I _think_ it's #289):
/retest

justinsb:
Thank you @CatherineF-dev 

/approve
/lgtm

PR #291, opened 2023-01-09
:sparkles: bump k8s from 1.25 to 1.26

camilamacedo86:

We need to bump the k8s version in this project to be able to support the new version in kubebuilder


justinsb:
Thanks for this @camilamacedo86 .... I'm thinking if we can get #292 in then we can probably be more version-independent on our code.

If we do have to add the new methods, that isn't really a blocker though either, we have a release-0.13 branch which corresponds to controller-runtime 0.13, so we can just treat our main branch as 0.14.

(We should probably also consider numbering that is more client-friendly, i.e. more aligned to k8s versioning!)

camilamacedo86:
@johnsonj  @justinsb @atoato88, we need to move forward with this one. 
**It is blocking kubebuilder**. 

See that the job pull-declarative-test is broken , check the open PRs in the repo. Example: https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/290

Could we force this one? 


@shane

camilamacedo86:
Hi @justinsb, 

Thank you a lot. It is rebased with the master now.
Currently we use the commits and not the branches there, but we 

> we can probably be more version-independent on our code.

In Kubebuilder we have now implemented: https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-2.md

That means we could move the declarative plugin to this repo which has its code. Therefore, we would release from here the plugin and not no longer be in Kubebuilder. However, Kubebuilder users would still be able to use it with the projects.  

That would decouple, and might help us keep a better quality of the plugin. Since it might get outdated many times because we are not synced with the latest changes. We can speak and look at that in the future. 

> If we do have to add the new methods, that isn't really a blocker though either, we have a release-0.13 branch which corresponds to controller-runtime 0.13, so we can just treat our main branch as 0.14.

We use commits there and not branches. So, when this PR get merged we are good to fly.

justinsb:
Thanks @camilamacedo86 - that's a great roadmap!

/approve
/lgtm

PR #297, opened 2023-01-27
make kustomize hydration a no-op

yuwenma:
This PR is aimed at downgrade the kustomize usage in k-d-p, which itself can potentially cause issues on imports and for other feature extension.

TODO: We'd like to use a build constraint to not deprecate methods. I'll work on that later once I get time.


yuwenma:
cc @justinsb

yuwenma:
Close this issue because I made a better approach in https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/299 (go build constraint). Many thanks to @justinsb for the recommendation!

PR #299, opened 2023-01-28
Opt-out the kustomize feature via go build tags 

yuwenma:
Problems:
- The k-d-p supports a kustomize hydration feature.  This feature requires library `kustomize/api`. This library itself is commonly used by other libraries and users can encounter dependency conflicts if the other library uses  a different `kustomize/api` release (what happened on my side and I cannot bump the other kustomize/api version).
- The k-d-p kustomize feature itself potentially causes friction issues, like [fs operation](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/ef14bb3642b4cb83bb5d93fe18ad7390665654c5/pkg/patterns/declarative/reconciler.go#L195), [run kustomize before other transformation operations](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/ef14bb3642b4cb83bb5d93fe18ad7390665654c5/pkg/patterns/declarative/reconciler.go#L414).

Description: 

This PR split the kustomize hydration to 1. a generic kustomize build and 2. a no-op in different files; It  adds a go build constraint `without_kustomize` to give users the ability to switch to opt out the kustomize feature in build time and no need to change their source code. 

Test:
This PR is tested using the guestbook example by [changing the `make run`](
https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/examples/guestbook-operator/Makefile#L63) for the following two cases

1. Preserve existing feature
```bash
make run
```

output 
```
2023-01-27T23:29:57-08:00	INFO	running kustomize	{"controller": "guestbook-controller", "object": {"name":"guestbook-sample","namespace":"default"}, "namespace": "default", "name": "guestbook-sample", "reconcileID": "fbcb7051-36c0-4358-aafc-5c84c6b96e2a", "kustomizer": "EnabledKustomize"}
```

2. Skip kustomize
```
# Makefile
run: manifests generate fmt vet ## Run a controller from your host.
	go run -tags without_kustomize ./main.go
``` 
```bash
make run
```

output 
```
2023-01-27T23:38:37-08:00	INFO	skip running kustomize	{"controller": "guestbook-controller", "object": {"name":"guestbook-sample","namespace":"default"}, "namespace": "default", "name": "guestbook-sample", "reconcileID": "38da0981-8143-43e5-96a4-7e551018758f", "kustomizer": "SkipKustomize"}
```

### UPDATED
```
go run -tags without_kustomize ./main.go

...
2023-01-28T23:16:16-08:00	ERROR	building deployment objects	{"controller": "guestbook-controller", "object": {"name":"guestbook-sample","namespace":"default"}, "namespace": "default", "name": "guestbook-sample", "reconcileID": "b62ee9fb-be37-4094-9bb7-c3cce756c03e", "error": "error running kustomize: kustomize support is not compiled in (built with tag `without_kustomize`)"}
```

```
go run ./main.go

2023-01-28T23:19:46-08:00	INFO	running kustomize	{"controller": "guestbook-controller", "object": {"name":"guestbook-sample","namespace":"default"}, "namespace": "default", "name": "guestbook-sample", "reconcileID": "eb7c1f04-771e-45a1-b44f-d1b451357d46", "manifestPath": "channels/packages/guestbook/0.0.1"}
```


pkg/patterns/declarative/kustomize/disabled.go
```diff
@@ -0,0 +1,27 @@
...
+var _ Kustomizer = &SkipKustomize{}
+
+// NoopKustomize ignore the `kustomization.yaml` file and won't run `kustomize build`.
+type SkipKustomize struct{}
+
+// Run is a no-op
+func (k *SkipKustomize) Run(ctx context.Context, _ filesys.FileSystem, _ string) ([]byte, error) {
+	log := log.FromContext(ctx)
+	log.WithValues("kustomizer", "SkipKustomize").Info("skip running kustomize")
+	return nil, nil
```
justinsb, line 26: I think you could return an error here, like `fmt.Errorf("kustomize support is not compiled in (built with tag without_kustomize)")`

pkg/patterns/declarative/kustomize/enabled.go
```diff
@@ -0,0 +1,38 @@
...
+	Kustomize = &EnabledKustomize{}
+}
+
+var _ Kustomizer = &EnabledKustomize{}
+
+// EnabledKustomize runs kustomize build. It requires kustomize/api v0.12.1 and above
+type EnabledKustomize struct{}
+
+// Run calls the kustomize/api library to run `kustomize build`.
+func (k *EnabledKustomize) Run(ctx context.Context, fs filesys.FileSystem, manifestPath string) ([]byte, error) {
```
justinsb, line 22: FYI you can have the same function with a build tag and without the build tag.

So you could declare `RunKustomize` in both.

One would have the build-tag `//go:build without_kustomize`, the other would have the build tag `//go:build !without_kustomize`.

Not a blocker though - and the refactoring is nice - but may be simpler in future.

yuwenma: Thank you for the suggestion! I'm trying to use function/struct name to make the code self-explained. But I agree the `EnabledKustomize` and `SkipKustomize` structs may give duplicate info since the go file name ("disabled" "enabled) and the build tags have already explained the rationale. Updated!

pkg/patterns/declarative/kustomize/disabled.go
yuwenma: I like the error message. Updated

justinsb:
Thanks @yuwenma!

/approve
/lgtm

atoato88:
LGTM :smile: 
Should we add documentation about `without_kustomize` tag?

yuwenma:
@atoato88 that's a good idea. I updated the doc here https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/301 Please let me know if there's anywhere else I shall update.

PR #300, opened 2023-01-29
WIP - KB declarative pattern external plugin

### LINKED ISSUE #293: Proposal on moving KB declarative pattern to external plugin
**What would you like to be added**:
This is a follow up based on the conversation in https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/291#issuecomment-1381123158. With the introduction of Phase 2 (out-of-tree external plugins) in Kubebuilder, we now have the flexibility to extend Kubebuilder and build/release any plugin independently outside the KB project. This paves way for smoother version upgrades across projects and eases the maintenance effort. 

Given this, we would like to know if we can go ahead with moving the Kubebuilder-declarative-pattern out of Kubebuilder, and building an external plugin for the same. The immediate benefits of this are:
- KB release need not be blocked on KB-declarative-pattern to upgrade dependencies during every K8s release cycle.
- The maintenance and release of KB-declarative-plugin can be handled independently, instead of relying on KB.

This would decouple the inter-dependency of both the projects.

em-r:
This PR copies the code of the declarative pattern plugin from [kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) to kubebuilder-declarative-pattern, as part of the ongoing effort to decouple the development of the plugin from the development of Kubebuilder now that Kubebuilder supports external plugins.

Fixes #293

Work done so far:
* Copied the code of the declarative pattern plugin from Kubebuilder.
* Install missing dependencies. `go mod tidy`
* Fix missing imports.

Remaining items:

- [ ] Discuss the current strategy. (e.g. Should we make the plugin a module?)
- [ ] Add an entry point (`main` function) since external plugins are shipped as binaries. ([docs](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/extensible-cli-and-scaffolding-plugins-phase-2.md))
- [ ] Write tests.
- [ ] Write documentation.

**Additional documentation**:


em-r:
cc @justinsb @camilamacedo86 
I've left a summary of the work done so far on the PR description.
Thanks!

em-r:
due to how Kubebuilder doesn't propagate project config on `init` command, but rather injects it in `PluginRequest.Universe` as part of `PROJECT` file (in subsequent commands e.g. `create api`), when externalizing declarative plugin, only one of the existing go plugins (e.g. `go/v4`, `go/v3`)  can be supported for performing scaffolding.
So to support scaffolds with different versions of go plugin, https://github.com/kubernetes-sigs/kubebuilder/issues/3245 needs to be resolved by adding support for injecting project config into external plugins, similar to how it's done for plugins that are builtin to Kubebuilder.
cc @camilamacedo86

camilamacedo86:
Hi @em-r,

The https://github.com/kubernetes-sigs/kubebuilder/issues/3245 is sorted out and available in the latest release 3.10. 
have we any other blocker to move forward within?

em-r:
Hi @camilamacedo86, thank you for following up on this.
I have just upgraded Kubebuilder locally to 3.10.0, however, I still couldn't find a way to access the project config or the layout via the input (e.g. PluginRequest) passed from Kubebuilder to the external plugin, could you please share with me how I can get access to one of those?

To recap: the problem is that the declarative plugin needs to know if the project is being scaffolded with `go/v4` or `go/v3` so that it can perform correct scaffolding around `Dockerfile` in the `init` subcommand. and unfortunately, when the `init` subcommand is running, the `PROJECT` file won't exist at that point so extracting the layout from it won't be an option, which leads me to think that we need to either "inject" the config (the way we do with internal plugins) through the `PluginRequest` input.

camilamacedo86:
Hi @em-r,

> [em-r](https://github.com/em-r) commented [last week](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/300#issuecomment-1529727708)
> Hi @camilamacedo86, thank you for following up on this.
> I have just upgraded Kubebuilder locally to 3.10.0, however, I still couldn't find a way to access the project config or the layout via the input (e.g. PluginRequest) passed from Kubebuilder to the external plugin, could you please share with me how I can get access to one of those?
> 
> To recap: the problem is that the declarative plugin needs to know if the project is being scaffolded with go/v4 or go/v3 so that it can perform correct scaffolding around Dockerfile in the init subcommand. and unfortunately, when the init subcommand is running, the PROJECT file won't exist at that point so extracting the layout from it won't be an option, which leads me to think that we need to either "inject" the config (the way we do with internal plugins) through the PluginRequest input. 


**The plugin go/v3 is deprecated. Therefore, I would strongly say that this plugin should only cares about go/v4 layout.** Therefore, the initial version should NOT care about or address this implementation. This complexity is sorted out. 

However, to improve the Kubebuilder API for future needs, I tracked the issue: https://github.com/kubernetes-sigs/kubebuilder/issues/3396

PR #307, opened 2023-02-15
housekeeping: update internal module references

justinsb:
We want to pick up the latest fixes to mockkubeapiserver.

We regenerate the test data also, we now have more API types.

Also we add more normalization around discovery requests, we try to reorder adjacent requests into alphabetical order, now that we are doing more of them this is more important for golden tests to be the same.


pkg/test/testreconciler/simpletest/testdata/reconcile/ssa/create/expected-http.yaml
```diff
@@ -108,40 +90,11 @@ Date: (removed)
 
 ---
 
-GET http://kube-apiserver/api/v1/namespaces/ns1/configmaps/foo
```
justinsb, line 111: Trying to figure out why we don't need to do this any more....

justinsb: Aha, I think it's a second reconciliation, and now we're doing it in one shot

justinsb:
/hold cancel

Changes look pretty good now, just more resources.

Will probably still be flaky until the _next_ PR goes in (starting watches before apply)

/retest

justinsb:
/retest

Took 4 tries previously, this is number 2

PR #308, opened 2023-02-17
Better `status.conditions`

yuwenma:
This PR is based on https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/272 which needs to go in first.

### Description
This PR adds a `status.conditions` according to the[ API convention "typical status properties"](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)  and the [kstatus abnormal-true conditions](https://github.com/kubernetes-sigs/cli-utils/tree/master/pkg/kstatus#conditions) 

The conditions logic is described as below:
 
- All manifest `.status.conditions` are augmented via kstatus to normalize its "type" as `Stalled` (kstatus abnormal-true), `Reconciling` (kstatus abnormal-true) or normal .  These augmented conditions  works together with the manifests' aggregated status. They will determine the `declarativeObject` present condition.

- The `declarativeObject` condition falls into 3 categories
    - **Ready**: This type is a backward compatible and read by kstatus as the resource is fully reconciled. If all manifests are reconciled, the `declarativeObject` will be in `Current` phase, with a single `Ready` condition.
    - **Reconciling**:  If no failed manifests, but manifests are in `Terminating` or `InProgress`,  the `declarativeObject` condition will be `Reconciling`. It will update existing `Reconciling` conditions and reset the `Stalled` conditions. 
    - **Stalled**: as long as one manifests has Failed status. The `declarativeObject` condition will be "Stalled". if `declarativeObject` already has "stalled" condition, the existing condition will be updated with the present `Message`, `Reason` (required) and `LastTransitionTime`, where `Message`  are pulled from the manifests conditions. 

### Examples
Example for case "All manifests are reconciled"
```yaml
kind: MyCR
...
status:
  conditions:
  - lastTransitionTime: "2023-02-17T03:56:20Z"
    message: all manifests are reconciled
    reason: Normal
    status: "False"
    type: Success
  healthy: true
  phase: Current
```  

Example for case "some manifests are InProgress or Terminating"
```yaml
kind: MyCR
status:
  conditions:
  - lastTransitionTime: "2023-02-17T03:51:28Z"
    message: |
         apps/v1, Kind=Deployment/argocd/argocd-repo-server:Deployment does not have minimum availability.
         apps/v1, Kind=Deployment/argocd/argocd-server:Deployment does not have minimum availability.
    reason: ContainAbnormalTrueConditions
    status: "True"
    type: Reconciling
  healthy: false
  phase: InProgress
```

Example for case "known error (AppliedFailed)"
```yaml
status:
  conditions:
  - lastTransitionTime: "2023-02-17T05:07:05Z"
    message: |
      apps/v1, Kind=StatefulSet/argocd/argocd-application-controller:Ready: 0/1
    reason: ContainAbnormalTrueConditions
    status: "True"
    type: Reconciling
  healthy: false
  phase: Applying
```

### Alternatives

Because [conditions themselves should not be state machines](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties), we don't really need to capture all manifests' conditions in one. But we can have each manifest's present abnormal-true condition to represent a `declarativeObject` condition, as shown below. 

```yaml
kind: MyCR
status:
  conditions:
  # deployment A
    - lastTransitionTime: "2023-02-17T05:23:40Z"
       message: Deployment does not have minimum availability.
       reason: MinimumReplicasUnavailable
       status: "False"
       type: Available
   # deployment B
    - lastTransitionTime: "2023-02-17T05:13:38Z"
       message: Deployment does not have minimum availability.
       reason: MinimumReplicasUnavailable
       status: "False"
       type: Available
  healthy: false
  phase: InProgress
```
The downside is that the `declarativeObject` itself's condition then can no longer be computed or augmented by kstatus correctly (it can no longer map the legacy type/status to k8s resources since the resource now is `MyCR`).


yuwenma:
cc @justinsb

yuwenma:
### Updates:

Per offline discussion with @justinsb and kstatus author @mortent, this PR is simplified to reflect the current status as a single condition with type "ready" and status `true/false`. The status `true` means all the manifests are reconciled, and `false` means the manifests are in progress (or in bad situation but k-d-p cannot determine whether it can be self-healed and reconciled later ).   

### Example

Example for case "some manifests are InProgress or Terminating"

```yaml
kind: MyCR
...
status:
  conditions:
  - lastTransitionTime: "2023-02-23T09:07:38Z"
    message: |
      apps/v1, Kind=StatefulSet/mycr/mycr-application-controller:Ready: 0/1
    reason: AbnormalTrueConditions
    status: "False"
    type: Ready
  healthy: false
  phase: InProgress
```

Example for case "All manifests are reconciled"

```yaml
kind: MyCR
...
status:
  conditions:
  - lastTransitionTime: "2023-02-23T09:17:57Z"
    message: all manifests are reconciled.
    reason: Normal
    status: "True"
    type: Ready
  healthy: false
  phase: Current
```

pkg/patterns/addon/pkg/status/conditions.go
```diff
@@ -9,19 +9,34 @@ import (
...
+func SetReady(commonStatus *addonsv1alpha1.CommonStatus, abnormalTrueConditions []status.Condition) {
+	setCondition(metav1.ConditionTrue, commonStatus, abnormalTrueConditions)
+}
+
+func setCondition(status metav1.ConditionStatus, commonStatus *addonsv1alpha1.CommonStatus, abnormalTrueConditions []status.Condition) {
+	reason, message := humanMessagefromConditions(abnormalTrueConditions)
+	commonStatus.Conditions = []*metav1.Condition{{
+		Status:             status,
+		Type:               ReadyType,
+		LastTransitionTime: metav1.Now(),
```
mortent, line 35: Does this correctly update the `LastTransitionTime` only when the actual value of the `Status` field changes? It might be useful to use the functions available in `k8s.io/apimachinery/pkg/api/meta` for updating the conditions.

pkg/patterns/addon/pkg/status/conditions.go
```diff
@@ -9,19 +9,34 @@ import (
 )
 
 const (
-	AbnormalReason = "ContainAbnormalTrueConditions"
+	AbnormalReason = "AbnormalTrueConditions"
```
mortent, line 12: It is not clear to me what this actually means. Abnormal-true polarity was a concept in a previous version of the Kubernetes API guidelines that suggested that conditions should be constructed such that a `True` value indicated an abnormal situation. The `Ready` condition that we are using now, is an example of a condition that doesn't follow this concept.

pkg/patterns/addon/pkg/status/conditions.go
yuwenma: Ah, I thought the transition refers to the present reconciliation cycle (where users can use to see when the object was last processed by the controller).  "the actual value of the Status field changes", that makes a lot of sense.

pkg/patterns/addon/pkg/status/conditions.go
yuwenma: How about `HaveAbnormalTrueManifests`? `Ready` condition is for the present object (which does not use the abnormal-true concept), but this `reason` field and its corresponding `message` field are for all the other manifests which this object manages. 

For those manifests, I do applying the abnormal-true polariy: I use kstatus to unify their conditions and pulled out all the abnormal-true conditions to build the value for `message`. So the meaning of `Ready` "false" means the current object has some manifests who are in abnormal-true condition.

yuwenma: Eventually I changed this AbnormalReason to `ManifestsNotReady`. 😃

pkg/patterns/addon/pkg/apis/v1alpha1/common_types.go
```diff
@@ -50,6 +50,9 @@ type CommonStatus struct {
 	Healthy bool     `json:"healthy"`
 	Errors  []string `json:"errors,omitempty"`
 	Phase   string   `json:"phase,omitempty"`
+	// Conditions follows the API specification "Conditions" properties.
+	// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
justinsb, line 55: So as this is, whenever controllers update and recompile they will immediately get these conditions added to their CRD.

If we did it automatically we would need a release note at the very least, and I'm not sure we should force this on people.  The other option is that we could make it more modular, and we could create something like Patches where users can mix-in the behaviours that they want.

(I'm also wondering if we should just use reflection, instead of requiring users to implement the interface.  I think if it's a well-known type it's not unreasonable, and it is more user-friendly.)

pkg/patterns/addon/pkg/status/conditions.go
```diff
@@ -0,0 +1,64 @@
...
+//		    apps/v1, Kind=Deployment/argocd/argocd-repo-server:Deployment does not have minimum availability.
+//		    apps/v1, Kind=Deployment/argocd/argocd-server:Deployment does not have minimum availability.
+func humanMessagefromConditions(conditions []status.Condition) (reason, message string) {
+	if len(conditions) == 0 {
+		return NormalReason, "all manifests are reconciled."
+	}
+	abnormalMessage := ""
+	for _, cond := range conditions {
+		if cond.Message != "" {
+			abnormalMessage += fmt.Sprintln(cond.Message)
```
justinsb, line 60: I would just do `abnormalMessage += cond.Message + "\n"`

The reason is that if there's a %s in cond.Message, it'll be treated as a formatting directive and things get messy.

The other option is `fmt.Sprintln("%s", cond.Message)`, but I personally think string concatenation is easier to read in that case.

pkg/patterns/addon/pkg/status/conditions.go
```diff
@@ -0,0 +1,64 @@
...
+// as long as one deployment manifest has an abnormal condition.
+// The "message" contains each abnormal condition's "reason" and "message".
+// e.g.
+//
+//	 conditions:
+//	 - reason: ManifestsNotReady
+//	   message: |-
+//		    apps/v1, Kind=Deployment/argocd/argocd-repo-server:Deployment does not have minimum availability.
+//		    apps/v1, Kind=Deployment/argocd/argocd-server:Deployment does not have minimum availability.
+func humanMessagefromConditions(conditions []status.Condition) (reason, message string) {
```
justinsb, line 53: I wonder if this gets easier if you return a status.Condition ?

pkg/patterns/addon/pkg/status/kstatus.go
```diff
@@ -44,21 +44,26 @@ func (k *kstatusAggregator) BuildStatus(ctx context.Context, info *declarative.S
 		}
 	}
 
+	// Abnormal-true refers to the present state condition where something unusual happens.
```
justinsb, line 47: So I think we are actually aggregating the conditions that are not healthy.  "abnormal-true" vs "abnormal-false" refers to whether the "happy" value is true or false, but I think what we're doing here is just collecting all the abnormal conditions.  So I might track them in `abnormalConditions []status.Condition` (or a map by GVKNN like `map[GVKNN][]status.Condition` if that is important)

pkg/patterns/addon/pkg/status/kstatus.go
```diff
@@ -44,21 +44,26 @@ func (k *kstatusAggregator) BuildStatus(ctx context.Context, info *declarative.S
...
+	var totalAbnormalTrueConditions []status.Condition
+
 	if shouldComputeHealthFromObjects {
 		statusMap := make(map[status.Status]bool)
 		for _, object := range info.Manifest.Items {
 			gvk := object.GroupVersionKind()
 			nn := object.NamespacedName()
 
-			log := log.WithValues("kind", gvk.Kind).WithValues("name", nn.Name).WithValues("namespace", nn.Namespace)
+			log.WithValues("kind", gvk.Kind).WithValues("name", nn.Name).WithValues("namespace", nn.Namespace)
```
justinsb, line 59: So this line doesn't actually log (IIUC).  It sets up a log context which includes the kind and namesapce and name in any messages that are logged.  It means we don't need to repeat this information in each of the log messages below.  So I think we do want `log := log...` (as before)

pkg/patterns/addon/pkg/status/kstatus.go
```diff
@@ -70,22 +75,66 @@ func (k *kstatusAggregator) BuildStatus(ctx context.Context, info *declarative.S
...
 	}
 
 	if err := utils.SetCommonStatus(info.Subject, currentStatus); err != nil {
 		return err
 	}
+	return nil
+}
+
+// getAbnormalTrueConditions calculates the abnormal-true conditions that best describe the current state of the deployment manifests.
+func getAbnormalTrueConditions(ctx context.Context, unstruct *unstructured.Unstructured) []status.Condition {
```
justinsb, line 100: I would just rename to `getAbnormalConditions`

pkg/patterns/addon/pkg/status/kstatus.go
```diff
@@ -70,22 +75,66 @@ func (k *kstatusAggregator) BuildStatus(ctx context.Context, info *declarative.S
...
+	log := log.FromContext(ctx)
```
justinsb, line 101: As another example, I think you could do

```
log := log.FromContext(ctx).WithValues("object", unstruct)
```

And then you would know that object would be in all your log messages in this function.

It's only two instances here, so it doesn't matter, but just wanted to try to clarify when you might use this!

pkg/patterns/addon/pkg/status/kstatus.go
```diff
@@ -70,22 +75,66 @@ func (k *kstatusAggregator) BuildStatus(ctx context.Context, info *declarative.S
...
+	if err != nil {
+		log.WithValues("object", unstruct).Error(err, "unable to get conditions")
+		return nil
+	}
+
+	if len(obj.Status.Conditions) == 0 {
+		return nil
+	}
+
+	// kstatus appends the augmented conditions to the end.
```
justinsb, line 121: Is there another method we can call into that is easier to use?

justinsb:
Some nits, looks really good.  I think / hope if you rebase the tests will be much less flaky

pkg/patterns/addon/pkg/status/conditions.go
yuwenma: Yeah it is. Thanks for the suggestion.

pkg/patterns/addon/pkg/status/kstatus.go
yuwenma: Unfortunately no. I read all `kstatus/status` code, and my gut feeling is that kstatus is trying to make the surfaced functions extremely simple and handles a huge amount of legacy and convention logic under the hood.

pkg/patterns/addon/pkg/status/kstatus.go
yuwenma: Agree. I renamed the variable and updated the comment. This code is in my first commit, I didn't update the name after our discussions.

pkg/patterns/addon/pkg/apis/v1alpha1/common_types.go
yuwenma: I changed the condition to use the patch-like approach and verified in my own operator: 
```
--> `all manifests are ready` 
--> delete a Deployment 
--> condition shows `ManifestsNotReady` 
--> wait for deployment to auto recover 
--> condition shows `all manfiests are ready` again . 
```
The change is in the last commit 
[`f4fd903` (#308)](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/308/commits/f4fd903209085de8188d5799c2646a6cf779433f)

pkg/patterns/addon/pkg/status/kstatus.go
```diff
@@ -44,21 +44,25 @@ func (k *kstatusAggregator) BuildStatus(ctx context.Context, info *declarative.S
...
+	var abnormalConditions []status.Condition
+
 	if shouldComputeHealthFromObjects {
 		statusMap := make(map[status.Status]bool)
 		for _, object := range info.Manifest.Items {
 			gvk := object.GroupVersionKind()
 			nn := object.NamespacedName()
 
-			log := log.WithValues("kind", gvk.Kind).WithValues("name", nn.Name).WithValues("namespace", nn.Namespace)
+			log = log.WithValues("kind", gvk.Kind).WithValues("name", nn.Name).WithValues("namespace", nn.Namespace)
```
justinsb, line 58: I think we need to keep the shadowing (i.e. `log := log.WithValues...`) ; we're creating a new log here, rather than assigning back to the original log value.  Once we exit the loop, log will no longer include `kind` / `name` / `namespace`. 

This code might be too subtle for its own good...  Sorry if I got this wrong before!

pkg/test/httprecorder/request_log.go
```diff
@@ -53,12 +57,34 @@ func (r *Request) FormatHTTP() string {
...
 	b.WriteString("\n")
 	if r.Body != "" {
-		b.WriteString(r.Body)
+		body := resetTimestamp(r.Body)
+		b.WriteString(body)
 		b.WriteString("\n\n")
 	}
 	return b.String()
 }
+func resetTimestamp(body string) string {
```
justinsb, line 66: This is a nice way to do things, but let's not do it here.  Instead can we move it a method like `ReplaceURLPrefix`?  The idea is that we capture the accurate log, and then deliberately apply a few normalizations, doing it that way should make debugging a bit easier because we can still see the "real" values.

The other way to do with time is to have a "Clock" interface, and have the default implementation return the time using `time.Now`, but have a version we use in tests that returns a controlled value.

The nice thing about your normalization approach is that we can more easily test against a real kubeapiserver also, so I like your approach better than injecting a mock Clock :-)

Let's just move the logic to a "fixer" function

justinsb (approved):

justinsb:
Two small things that we should fix, but this looks good and I am happy for you to fix those things in a separate PR if that is easier.  So

/approve
/lgtm

/hold

You should cancel the hold yourself if you agree and want to address the issues in a follow-on PR, but if you want to push a fix into this PR that works too.

pkg/test/httprecorder/request_log.go
yuwenma: you are right. I didn't think carefully about the test maintenance cost. Totally agree it could be hard to debug if the request/response body is modified and the logic is buried somewhere not directly reflected in the `_test.go` code.
Updated

pkg/patterns/addon/pkg/status/kstatus.go
yuwenma: Fixed. Thank you so much for pointing it out.

PR #309, opened 2023-02-18
Leverage kstatus for applyset healthcheck

yuwenma:
### Problem 
The current applyset relies on the `status.conditions` to determine a resource healthy condition. But some legacy k8s resources do not have `status.conditions`, like   `StatefulSet` `Service` `NetworkPolicy`. This results in `InternalError` even if those resources are actually healthy. 

As suggested in the `TODO` comment, kstatus provides a reliable approach on determining the status of those core k8s resources even if they have different `status` specifications. 

However, kstatus is under cli-utils (originally in kustomize) and we do not want to import cli-utils in applyset. The workaround in this PR is to use a specific kstatus version when it was a standalone go library in kustomize. This introduces the minimum dependencies.

related go libraries:
kstatus is removed from kustomize in api/v0.4.2 https://github.com/kubernetes-sigs/kustomize/pull/2674
this PR uses the version from api/v0.4.1


go.mod
```diff
@@ -16,12 +16,12 @@ require (
 	github.com/prometheus/client_golang v1.14.0
 	golang.org/x/crypto v0.1.0
 	golang.org/x/tools v0.3.0
-	k8s.io/api v0.26.0
-	k8s.io/apimachinery v0.26.0
-	k8s.io/cli-runtime v0.26.0
-	k8s.io/client-go v0.26.0
+	k8s.io/api v0.26.1
+	k8s.io/apimachinery v0.26.1
+	k8s.io/cli-runtime v0.26.1
```
barney-s, line 21: @justinsb - Would it be possible to remove these dependencies (added via kstatus) 
This pulls in kubectl aswell.
Looking for a trimmed import list.

go.mod
```diff
@@ -16,12 +16,12 @@ require (
...
+	k8s.io/client-go v0.26.1
 	k8s.io/klog/v2 v2.80.1
-	k8s.io/kubectl v0.26.0
+	k8s.io/kubectl v0.26.1
```
barney-s, line 24: kstatus -> cli-utils -> cli-runtime -> kubectl ?

barney-s (commented):
Would it be possible to trim some dependencies ?

go.mod
yuwenma: Yes, I think that's the plan. The kstatus methods used here do not depend on `kubectl`.  It's the other methods that introduces the `kubectl`. So remove the `kubectl` dependency may require some kstatus pkg refactoring (not sure how feasible that is). Or we can fork the kstatus in this k-d-p repo.   

I probably should move this PR back to the WIP status, since I'm not planning to work on the applyset yet (I was on some other stuff related to kstatus and `status.condition` and I saw the TODO comment in this file about using kstatus)

applylib/applyset/health.go
```diff
@@ -17,105 +17,34 @@ limitations under the License.
...
 )
 
 // isHealthy reports whether the object should be considered "healthy"
 // TODO: Replace with kstatus library
 func isHealthy(u *unstructured.Unstructured) bool {
-	// Check if the resource is scheduled for deletion
-	deletionTimestamp := u.GetDeletionTimestamp()
-	if deletionTimestamp != nil {
-		klog.Infof("object %s is scheduled for deletion", humanName(u))
+	result, err := status.Compute(u)
```
yuwenma, line 30: `status.Compute` contains the logics from the deleted codes (deletionTimestamp, no-condition k8s resources and legacy status.conditions field).

applylib/go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	k8s.io/klog/v2 v2.90.1
 	sigs.k8s.io/controller-runtime v0.14.1
 	sigs.k8s.io/kubebuilder-declarative-pattern/mockkubeapiserver v0.0.0-20221021151406-9bd3fb842119
+	sigs.k8s.io/kustomize/kstatus v0.0.2-0.20200509233124-065f70705d4d
```
yuwenma, line 17: kstatus is under cli-utils (originally in kustomize) and we do not want to import cli-utils in applyset. This workaround  is to use the kstatus when it was a standalone go library in kustomize. This introduces the minimum dependencies.

Related libraries:
kstatus is removed from kustomize in api/v0.4.2 https://github.com/kubernetes-sigs/kustomize/pull/2674
this PR uses the commit from api/v0.4.1

applylib/applyset/health.go
```diff
@@ -17,105 +17,34 @@ limitations under the License.
...
+	case status.InProgressStatus:
+		return false
+	case status.FailedStatus:
+		return false
+	case status.TerminatingStatus:
+		return false
+	case status.UnknownStatus:
+		klog.Warningf("unknown status for %s", humanName(u))
+		return false
+	default: // status.CurrentStatus:
```
justinsb, line 45: Nit: Should we affirmatively match status.CurrentStatus?  (And then log a similar warning to UnknownStatus for default)

applylib/go.mod
justinsb: Seems reasonable!

justinsb (approved):

justinsb:
Thanks @yuwenma 

/approve
/lgtm

applylib/applyset/health.go
yuwenma: Sure/ here's the PR https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/329

PR #311, opened 2023-03-01
mockkubeapiserver: Always send content-length

justinsb:
This allows for more predictable output, as otherwise the length is
only sometimes sent (maybe varying by go version)

### DIFF
--- a/mockkubeapiserver/request.go
+++ b/mockkubeapiserver/request.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strconv"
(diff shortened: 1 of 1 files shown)


yuwenma (approved):
Looks great

atoato88:
I know that this PR already has lgtm label and be merged, but it looks good from me too.

justinsb:
Thank you @yuwenma and @atoato88  :-)

Trying to get these tests to pass reliably - I think i know what the problem is - I think we should be setting up the watches before even starting the apply, so I'm working towards that!

PR #315, opened 2023-03-03
kubectl applier: use an empty cache dir in testing

justinsb:
This ensures that we don't reuse the discovery data from a previous
run, if we happen to land on the same port etc.


yuwenma (approved):
This is cool!

justinsb:
I haven't seen this since, so I'm going to close.  If we need it, it's always here!

/close

PR #316, opened 2023-03-03
Simplify List flattening logic

justinsb:
Refactor to ensure that we build and flatten the objects in one place,
simplifying execution.  The intermediate call to VersionCheck (should)
prefer flattened objects.


atoato88:
It looks that updating the test case is needed from [this result](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kubebuilder-declarative-pattern/316/pull-declarative-test/1632952839514886144#1:build-log.txt%3A608-699).

Is it caused by updating target object with flatten one?

atoato88:
The test result seems flaky one? :thinking:

yuwenma (approved):

PR #318, opened 2023-03-16
reset status.healthy in each reconcile

yuwenma:

The k-d-p  object should set `status.healthy=true` if all manifests are reconciled. Right now if the previous reconcile is unhealthy, this field will show "healthy=false" even if the latest reconcile is healthy.

### DIFF
--- a/pkg/patterns/addon/pkg/status/kstatus.go
+++ b/pkg/patterns/addon/pkg/status/kstatus.go
@@ -7,7 +7,6 @@ import (
 	"sigs.k8s.io/cli-utils/pkg/kstatus/status"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
-	addonsv1alpha1 "sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/addon/pkg/apis/v1alpha1"
 	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/addon/pkg/utils"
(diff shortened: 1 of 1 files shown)


yuwenma:
/hold

add testing coverage

pkg/patterns/addon/pkg/status/kstatus.go
```diff
@@ -88,7 +87,7 @@ func (k *kstatusAggregator) BuildStatus(ctx context.Context, info *declarative.S
 		}
 		currentStatus.Phase = string(aggregatedPhase)
 	}
-
+	currentStatus.Healthy = currentStatus.Phase == string(status.CurrentStatus)
```
yuwenma, line 90: I'd like to confirm if the logic is correct, basically it means 
1. reconcile does not have errors like `InternalError`,  `VersionMismatch`, `FailedToApply` 
2. reconcile is not in progress, including `Terminating` and `InProgress`
3. reconcile is not failed.
5. all deployment manifests are reconciled, and no abnormal conditions are detected (this is guaranteed from the way how we calculate the `current` status)

If this logic is reasonable, the `status.healthy` and `status.phase` seems to be duplicate fields. I think it's fine since [`phase` is deprecated](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md), just want to double check this is what k-d-p wants.

justinsb:
Thanks  - nice catch.  Can you squash please though?

/approve
/lgtm

/hold for squash

justinsb:
/hold cancel

Maybe the robots can do it!

PR #319, opened 2023-03-17
make status.condition patchable

yuwenma:
This change improves the status.conditions for the following 3 scenarios:

#### Current behavior
If users update the k-d-p in their operator, they have to regenerate the CRD because by default the api schema contains the `status.conditions` field in the default `commonstatus`. users don't need to change their api schema. 

#### New behavior
If users do not want to add the patchable conditions, after they update the k-d-p they can keep their existing behavior and the CR object will not have `.status.conditions` field.

### Case 2: Default `status.conditions`   
users can use the new `status.conditions` simply by adding the `PatchableConditions` to their CR obejcts' `Status` api (see below). They should re-generate the CRD and DeepCopy objects to contain new `status.conditions`, but they don't need to implement the `addonsv1alpha1.ConditionGetterSetter` interface.

```go
type MyObjStatus struct {
	addonv1alpha1.CommonStatus        `json:",inline"`
	addonv1alpha1.PatchableConditions `json:",inline"` # NEW
}

type MyObj struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   MyObjSpec   `json:"spec,omitempty"`
	Status MyObjStatus `json:"status,omitempty"`
}
```

### Case 3: Customize the `status.conditions`   

If users add the patchable conditions and want to implement their own condition setter/getter  (`addonsv1alpha1.ConditionGetterSetter`) to their declarative object, they should expect the customize getter/setter being used.

```go

type MyObjStatus struct {
	addonv1alpha1.CommonStatus        `json:",inline"`
	addonv1alpha1.PatchableConditions `json:",inline"` # NEW
}

var _ addonsv1alpha1.ConditionGetterSetter = &MyObjStatus{}
func (o *MyObjStatus) SetConditions(addonv1alpha1.PatchableConditions) {...}
func (o *MyObjStatus) GetConditions() addonv1alpha1.PatchableConditions {...}
```

TODO: Add the patchableConditions to guestbook example.


pkg/test/testreconciler/simpletest/controller_test.go
```diff
@@ -138,7 +138,7 @@ func testSimpleReconciler(h *testharness.Harness, testdir string, applier applie
 
 	mgrContext, mgrStop := context.WithCancel(ctx)
 	go func() {
-		time.Sleep(2 * time.Second)
+		time.Sleep(1 * time.Second)
```
yuwenma, line 141: With 2 second reconcile window, the test always ends up with several more roundtrips and this results as more `GET ..configmap, namespace` `PATCH configmap namespace`. Since all HTTP requests are logged and compared by golden test, the e2e test always fails with 3000+ lines added to the simpletest/testdata/reconcile/ssa/create/expected-http.yaml, than expected. 

Reducing the time window is a tmp workaround. Another option we can do is to skip logging the http request for "GET configmap/namespace", and "PATCH if no change"  (current e2e test logs the PATCH even if the manifest is not changed. https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/mockkubeapiserver/patchresource.go#L122-L124

justinsb: Ideally we would "know" that reconciliation is done, but that's actually a bit tricky to know.  A longer window ensures that there's no ongoing reconciliation, a shorter window means the tests run faster.

The "real" fix here I think is to ensure that deployments go ready, which I think is why we are looping here.  We're doing that separately.  I think reducing the time is fine here, we can revisit in future once all the in-flight PRs have landed.

justinsb (approved):

PR #320, opened 2023-03-20
mockkubeapiserver: deployments should become healthy

justinsb:
Write some simple stub-logic so that deployments immediately
have a status indicating that they are ready.

### DIFF
--- a/applylib/go.mod
+++ b/applylib/go.mod
@@ -8,9 +8,9 @@ go 1.19
 
 require (
 	github.com/google/go-cmp v0.5.9
-	k8s.io/api v0.26.0
+	k8s.io/api v0.26.2
 	k8s.io/apimachinery v0.26.2
-	k8s.io/client-go v0.26.0
+	k8s.io/client-go v0.26.2
 	k8s.io/klog/v2 v2.90.1
 	sigs.k8s.io/controller-runtime v0.14.1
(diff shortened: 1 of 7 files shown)


yuwenma (approved):
Thanks for the change. I like this approach better than making tricks around the reconcile duration.

atoato88:
/lgtm
I also agree with this update.

atoato88:
/retest
Can we pass the test? :thinking:

yuwenma:
> /retest Can we pass the test? 🤔

The e2e test is bundled with [a specific mock-kubeapiserver](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/go.mod#L28), so this PR won't help until it's merged and then we bump the mockkubeapiserver in the main k-d-p. Justin has another improvement on having the e2e test [using the latest mockkubeapiserver](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/321), which should allow the CI using the latest mockkubeapiserver, but itself is blocked by the flaky results that this PR supposes to help 😅

justinsb:
I'm trying to get the tests to pass by cherry-picking #321 onto this branch, so that we can see them pass together and merge them as one.  (#321 has lgtm so would merge if tests passed, but it would be good to prove that the combination here does indeed pass tests!)

justinsb:
/test pull-declarative-test

This is the _kubectl_ discovery being non-deterministic.  I think that's the remaining big source of non-determinism.  I had a look at the code this morning and I didn't see a problem ... but I'm going to keep looking.

PR #321, opened 2023-03-21
Run CI with go workspace

justinsb:
Because of the challenges of updating multiple modules at the
same time, we run our tests with a go workspace, so that
we test with all the latest code.

Maybe we should commit a go workspace to the repo, but conventional
advice (I believe) is not to do that.

Fixes #

**Additional documentation**:

### DIFF
--- a/applylib/go.mod
+++ b/applylib/go.mod
@@ -8,9 +8,9 @@ go 1.19
 
 require (
 	github.com/google/go-cmp v0.5.9
-	k8s.io/api v0.26.0
+	k8s.io/api v0.26.2
 	k8s.io/apimachinery v0.26.2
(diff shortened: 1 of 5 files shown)


yuwenma:
/lgtm

Considering that we have several PRs that improve the main k-d-p e2e test but the change is in the dependencies (applier and mock-kubeapiserver), making this CI change can greatly improve and expedite the developing cycle.

yuwenma:
Seems like the PRs are blocking each other (maybe merging https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/320 into this one? or merge with CI failure)

justinsb:
Done as part of #320 (had to merge both to get good test signal)

/close

PR #322, opened 2023-03-22
testreconciler: pause after reconciliation

justinsb:
We sleep briefly (100ms) after updating the status.

This is a hack to ensure that we see the watch event, because
otherwise we can start a re-reconcile based on our derived objects,
and then reconcile again based on our own status update.

This is racy behaviour and causes non-determinism which breaks our
golden tests.

We do this only in this test controller, but maybe we should have
similar logic to "debounce" in all controllers.

Fixes #

**Additional documentation**:

### DIFF
--- a/pkg/patterns/declarative/hooks.go
+++ b/pkg/patterns/declarative/hooks.go
@@ -48,3 +48,19 @@ type AfterApply interface {
 type BeforeApply interface {
 	BeforeApply(ctx context.Context, op *ApplyOperation) error
 }
+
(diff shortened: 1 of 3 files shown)


justinsb:
/test pull-declarative-test

We're getting closer, but sometimes kubelet seems to skip one or more discovery requests...

yuwenma (approved):
This makes a lot of sense. Besides`declarativeObject`, do we want to pass the to-update or updated manifests to the hook methods?

PR #323, opened 2023-03-23
controllerrestmapper: implement RESTMappings

justinsb:
This function is needed for mapping from a group kind
to a group version kind, which will be needed for pruning.


yuwenma:
/lgtm


Great pointers. Thank you so much!

PR #324, opened 2023-03-23
tests: fix race when recording concurrent requests

justinsb:
When kubectl was doing discovery in parallel, we would often drop
requests from the log.  Add a mutex to protect against this (and
generally to make it more correct)


yuwenma (approved):
I'm excited to see this change goes in!

atoato88:
We got stable test results thanks to this PR :smile:

PR #325, opened 2023-03-24
Prune logic for ApplySetApplier

yuwenma:
This PR adds the server-side pruning logic for ApplysetApplier. 
 
## How to use 

```go
	applier := applier.NewApplySetApplier(metav1.PatchOptions{}, metav1.DeleteOptions{}, applier.ApplysetOptions{}) 
	r.Reconciler.Init(mgr, &addonsv1alpha1.YourCR{},
		declarative.WithApplier(applier),
		declarative.WithApplyPrune())
```

### First Time Reconcile

* Your declarative CR object will have a new label `applyset.kubernetes.io/id: PARENT_ID` 
* All the deployment manifests will have a new label `applyset.kubernetes.io/part-of: PARENT_ID`
* Your declarative CR object will have annotations:
   * `applyset.kubernetes.io/contains-group-resources` which contains all the manifests GR
   * `applyset.kubernetes.io/additional-namespaces` which contains all the namespaces those manifests are in.
   * `applyset.kubernetes.io/tooling` whose default value is the your CR object's kind.
* All these labels and annotations are added even if the prune is not enabled.**

```yaml
# subject
apiVersion: addons.configdelivery.anthos.io/v1alpha1
kind: MyCR
metadata:
  annotations:
    applyset.kubernetes.io/additional-namespaces: mycr
    applyset.kubernetes.io/contains-group-resources: clusterrolebindings.rbac.authorization.k8s.io,clusterroles.rbac.authorization.k8s.io,configmaps,customresourcedefinitions.apiextensions.k8s.io,deployments.apps,namespaces,networkpolicies.networking.k8s.io,rolebindings.rbac.authorization.k8s.io,roles.rbac.authorization.k8s.io,secrets,serviceaccounts,services,statefulsets.apps
    applyset.kubernetes.io/tooling: MyCR/
  labels:
    applyset.kubernetes.io/id: applyset-yHSTf9d_-Ozn5cvv-JpOaK-5-W79ncyZpBVXt1_nZzc-v1
  name: mycr-sample
spec:
  version: 2.5.11

# a deployment in the manifest
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    addons.configdelivery.anthos.io/argocd: mycr-sample
    app.kubernetes.io/component: controller
    app.kubernetes.io/name: mycr-controller
    app.kubernetes.io/part-of: mycr
    applyset.kubernetes.io/part-of: applyset-yHSTf9d_-Ozn5cvv-JpOaK-5-W79ncyZpBVXt1_nZzc-v1
```

### New Manifest GVR

If you upgrade the declarative CR version, which contains new manifests GroupVersionResource, you can expect:
* `applyset.kubernetes.io/contains-group-resources` will include new GVR
* the new manifest will have new label `applyset.kubernetes.io/part-of`
* Under the hood, the new restmapping is also cached.

### Manifest no longer needed

If you upgrade the declarative CR version, which no longer contain certain manifests, you can expect:
* That manifest will be pruned
* If this is the single GVR type, your declarative CR `applyset.kubernetes.io/contains-group-resources` will be updated.

You can find logs like below
```
I0329 20:53:39.968892   57451 applyset.go:206] prune is enabled
I0329 20:53:40.282866   57451 applyset.go:330] pruned resource Deployment.apps <NS>/<DEPLOYMENT NAME>
I0329 20:53:40.282953   57451 applyset.go:225] prune succeed
```


applylib/applyset/results.go
```diff
@@ -27,6 +27,8 @@ type ApplyResults struct {
 	total             int
 	applySuccessCount int
 	applyFailCount    int
+	pruneSuccessCount int
+	pruneFailCount    int
```
justinsb, line 31: We probably need to include pruneFailCount in AllApplied ... AllApplied is sort of considered a "did we apply everything OK", and if we are now pruning as part of apply, we want that to be false if prune failed, so that we can surface the information that we aren't completely done yet.  (?)

applylib/applyset/results.go
```diff
@@ -49,6 +51,7 @@ func (r *ApplyResults) AllHealthy() bool {
 
 // checkInvariants is an internal function that warns if the object doesn't match the expected invariants.
 func (r *ApplyResults) checkInvariants() {
+	// TODO: Check the pruning results.
```
justinsb, line 54: TBH this is less important.  If you can think of a good sanity-check on the prune counters, great, but as this depends on the prior state I don't think there is one.

pkg/patterns/declarative/pkg/applier/applylib.go
```diff
@@ -29,9 +38,11 @@ func (a *ApplySetApplier) Apply(ctx context.Context, opt ApplierOptions) error {
 		switch arg {
 		case "--force":
 			opt.Force = true
-
+		case "--prune":
+			opt.Prune = true
 		default:
-			return fmt.Errorf("extraArg %q is not supported by the ApplySetApplier", arg)
+			// TODO(yuwenma): This skips the "--selector addons.configdelivery.anthos.io/MyCR=MyCRName"
+			klog.Errorf("extraArg %q is not supported by the ApplySetApplier", arg)
```
justinsb, line 45: I think you should specifically handle `--selector` and ignore it (perhaps with a warning), and keep the error on anything we don't know about.

pkg/patterns/declarative/pkg/applier/applylib.go
```diff
@@ -43,26 +54,34 @@ func (a *ApplySetApplier) Apply(ctx context.Context, opt ApplierOptions) error {
 	}
 
 	restMapper := opt.RESTMapper
+	tooling := a.Tooling
+	if tooling == "" {
+		tooling = opt.ParentRef.GroupVersionKind().Kind
```
justinsb, line 59: Ah - cool default - this should work well :-)

pkg/patterns/declarative/pkg/applier/applylib.go
```diff
@@ -43,26 +54,34 @@ func (a *ApplySetApplier) Apply(ctx context.Context, opt ApplierOptions) error {
...
 	}
 	s, err := applyset.New(options)
 	if err != nil {
 		return fmt.Errorf("error creating applyset: %w", err)
 	}
 
-	// Populate the namespace on any namespace-scoped objects
-	if opt.Namespace != "" {
-		for _, obj := range opt.Objects {
-			gvk := obj.GroupVersionKind()
```
justinsb, line 60: Lol - looks like I assumed restMapper cached as well ;-)

Any rason to move this out of the loop?  Are you trying to catch errors where the GVK is not yet known?  I don't _think_ you want to error out here, I think the correct answer is probably a partial apply (with correct management of the resources-in-use annotation in the parent)

pkg/patterns/declarative/reconciler.go
```diff
@@ -232,7 +233,11 @@ func (r *Reconciler) updateStatus(ctx context.Context, original DeclarativeObjec
 	}
 
 	if !reflect.DeepEqual(oldStatus, newStatus) {
-		if err := r.client.Status().Update(ctx, instance); err != nil {
+		data, err := json.Marshal(newStatus)
+		if err != nil {
+			log.Error(err, "error marshaling new status %v: %v", newStatus, err)
```
justinsb, line 238: return an error I think

pkg/patterns/declarative/reconciler.go
```diff
@@ -232,7 +233,11 @@ func (r *Reconciler) updateStatus(ctx context.Context, original DeclarativeObjec
...
+		}
+		if err := r.client.Status().Patch(ctx, instance, client.RawPatch(types.ApplyPatchType, data)); err != nil {
```
justinsb, line 240: Good change but ... is it related / required by this PR?

pkg/patterns/declarative/reconciler.go
```diff
@@ -359,10 +364,16 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 		}
 	}
 
+	// TODO(yuwenma): instead of passing in the applyset "parent" as a applier option, maybe it's better to pass in the
```
justinsb, line 367: I think to keep controller-runtime happy you do want to be working with the same instance of the subject.  It writes back the changed resourceversion I believe, so you can call update twice.

pkg/restmapper/caching.go
```diff
@@ -70,12 +72,35 @@ type cachedGVR struct {
 	Scope    meta.RESTScope
 }
 
+// KindFromGVR finds out the Kind from the GVR in the cache. If the GVR version is not given, we will iterate all the matching
+// GR in the cache and return the first matching one.
+// e.g. https://github.com/kubernetes/kubernetes/blob/a0cff30104ea950a5cc733a109e7f9084275e49e/staging/src/k8s.io/kubectl/pkg/cmd/apply/applyset.go#L353
+func (c *cache) KindFromGVR(gvr schema.GroupVersionResource) string {
```
justinsb, line 78: I think this is the same function as RESTMappings ?  Do we need this?

pkg/restmapper/controllerrestmapper.go
```diff
@@ -56,7 +60,22 @@ func (m *ControllerRESTMapper) ResourcesFor(input schema.GroupVersionResource) (
 // RESTMapping identifies a preferred resource mapping for the provided group kind.
 func (m *ControllerRESTMapper) RESTMapping(gk schema.GroupKind, versions ...string) (*meta.RESTMapping, error) {
 	ctx := context.TODO()
-
+	// Since versions is optional string slice, it can be empty. If version is not given, we will iterate all the cached
+	// GV and find the first matching RESTMapping.
+	if len(versions) == 0 {
+		for keyGV, cachedGV := range m.cache.groupVersions {
```
justinsb, line 66: I think you can delegate this case to a call to RESTMappings ?

pkg/test/testreconciler/simpletest/testdata/reconcile/direct/create/expected-http.yaml
```diff
@@ -114,6 +114,32 @@ Date: (removed)
 
 ---
 
+GET http://kube-apiserver/api/v1?timeout=32s
```
justinsb, line 117: Not entirely sure why this one is changing...  Maybe try updating the golden output with a workspace?

yuwenma:
@justinsb Regarding our discussions about RESTMappings, I think it involves two problems (see the third commit) and it's kind of complicated in the code, so I'd like to explain a little bit more:


### Problem 1 

CRD and its CR objects are deployed in reverse order and it results in missing RESTMapping 
 
#### Solution

We uses the `k-d-p/ControllerRESTMapper` to keep track and cache all the RESTMappings in history (ControllerRESTMapper is [initialized on the k-d-p/Reconciler level](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/325/files#diff-91d9000f2817ed2698caf7b294e34ed1bc739bf60b7dc1af3136055b18741312R122)). 

During the applyset `Apply`, If one manifest apply fails, we [continue with the next manifest](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/325/files#diff-a66647ca49c9a35b9a09531b4095cbcae95235d05f4a18a12b01f904ba9bb8e8R195). The failed apply will [trigger a re-deploy](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/325/files#diff-550c225bbd4c46cb6a248d12223e3a936c129daa9112ed7bb8e5c97ad8b3c797R117-R118), so the CR object can be applied successfully. And both CRD and CR RESTMappings are cached in the `ControllerRESTMapper`.  

This guarantees that if the CR object is ordered before CRD, both of them can still be deployed successfully in two `Apply` runs, and both RESTMappings are cached. 

### Problem 2

`k-d-p/ applyset` needs to provide `kubectl.ApplySet` the RESTMappings info so that it can efficiently find the right objects to prune. 

This actually involves **two different kinds of RESTMappings**
1. The RESTMappings from last deployed manifests. This RESMappings is used to parse the parent's  `applyset.kubernetes.io/contains-group-resources` annotation to get the previously applied manifests GVRs.
1.  The current manifests RESTMappings to determine the up-to-date GVRs. 

#### Solution

Besides caching all the RESTMappings as an attribute to `ApplysetApplier`, we initialize the `kubectl ApplySet` [in each `Apply` run](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/325/files#diff-a66647ca49c9a35b9a09531b4095cbcae95235d05f4a18a12b01f904ba9bb8e8R137), and we provide both caching-all RESTMapping and current-only RESTMapping separately to the `kubectl ApplySet` 

1. The `kubectl.ApplySet` will use the [caching-all RESTMapper to parse the parent annotation](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/325/files#diff-aa80169bbcde9041cc2ddeeafb2d94d32325bea91e91f7c129d89da0e4ab24a5R286) to get the previously applied GVRs.
1. k-d-p `ApplySet` will [register the current manifest's RESTMappings](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/325/files#diff-a66647ca49c9a35b9a09531b4095cbcae95235d05f4a18a12b01f904ba9bb8e8R154) after it's successfully applied to the `kubectlapply.ApplySet` (a different var than the caching-all RESTMappings)

yuwenma:
@justinsb  Besides the RESTMapping, another problem is about the tradeoff between the reliability and efficiency of updating the parent object in the cluster. I'd like to know if the following change makes sense to you:

#### The previous workflow is 
1. ApplysetApplier [caches the current manifests' RESTMappings](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/325/commits/030d8409bc04d1d97e766eee35223301b9678e58#r1154958822)
1. `ApplyOnce` updates the parent (**first call, `Patch`**) labels and annotations with the RESTMappings infos (requires both the current and caching-all, since it is a "superset"), this is the `BeforeApply` method
1. After apply, updates the parent  (**second call, `Patch`**) labels and annotations with the RESTMappings info (requires the current-only RESTMappings, this is a "last") 
1. In the main Reconcile, update the parent status (**third call, `Status.Patch`**)

#### Problems
The problem of the previous workflow is that we cannot get the current manifests RESTMappings before apply efficiently, I think you [pointed it out](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/325/commits/030d8409bc04d1d97e766eee35223301b9678e58#r1154882942) here but we cannot return errors either because it will end up the `ApplyOnce` with no-op. 

#### Trade off 

I remove the `BeforeApply` completely. The downside is that if the operator cashes during manifests apply, the parent may not be able to track the current manifest GVRs, and this can end up as missed pruning objects. 
  
* The new workflow is 
1. `ApplyOnce` applies manifest, and then caches each manifest's RESTMapping for `kubectl.ApplySet` if the apply succeeds.
2. Parent is passed to `ApplySet` as `runtime.Object`, and the labels and annotations is updated at the end of the pruning. 
3. The Parent is updated in the main `Reconcile` after apply succeeds.  (**first call, `Patch`**) 
4. The parent status is updated at last (**second call, `Status.Patch`**)

applylib/applyset/applyset.go
```diff
@@ -100,6 +141,43 @@ func (a *ApplySet) ApplyOnce(ctx context.Context) (*ApplyResults, error) {
...
+		// TODO: Check error is NotFound and not some transient error?
+		restMapping := result.restMapping
+		if restMapping != nil {
+			// cache the GVK in kubectlapply. kubectlapply will use this to calculate
+			// the latest parent "applyset.kubernetes.io/contains-group-resources" annotation after applying.
+			kapplyset.AddResource(restMapping, obj.GetNamespace())
+		}
+	}
+	if err := a.WithParent(ctx, kapplyset); err != nil {
+		return results, fmt.Errorf("unable to update Parent %v", err)
```
justinsb, line 179: Nit: `"...: %w"`

applylib/applyset/applyset.go
```diff
@@ -110,11 +188,23 @@ func (a *ApplySet) ApplyOnce(ctx context.Context) (*ApplyResults, error) {
...
+		if restMappingResult.err != nil {
+			results.applyError(gvk, nn, fmt.Errorf("error getting rest mapping for %v: %w", gvk, restMappingResult.err))
 			continue
 		}
+
+		restMapping := restMappingResult.restMapping
+		if restMapping == nil {
+			// Should be impossible
+			results.applyError(gvk, nn, fmt.Errorf("rest mapping result not found for %v", gvk))
+		}
```
justinsb, line 201: Ooops ... I forgot a `continue` here!

applylib/applyset/applyset.go
```diff
@@ -110,11 +188,23 @@ func (a *ApplySet) ApplyOnce(ctx context.Context) (*ApplyResults, error) {
...
+
+		if err := a.updateManifestLabel(obj, kapplyset.LabelsForMember()); err != nil {
+			klog.Errorf("unable to update label for %v/%v %v: %v", obj.GetName(), obj.GetNamespace(), gvk, err)
```
justinsb, line 204: `return fmt.Errorf` , I think, otherwise we will leak the object (?)

applylib/applyset/applyset.go
```diff
@@ -153,11 +242,151 @@ func (a *ApplySet) ApplyOnce(ctx context.Context) (*ApplyResults, error) {
...
+				return err
+			}
+			if err = a.deleteObjects(ctx, pruneObjects, results); err != nil {
+				return err
+			}
+			return nil
+		}()
+		if err != nil {
+			klog.Errorf("encounter error on pruning %v", err)
+			if e := a.updateParentLabelsAndAnnotations(ctx, kapplyset, "superset"); e != nil {
```
justinsb, line 267: I don't think we need to update the parent labels here?

applylib/applyset/applyset.go
```diff
@@ -153,11 +242,151 @@ func (a *ApplySet) ApplyOnce(ctx context.Context) (*ApplyResults, error) {
...
+				return results, e
+			}
+		}
+	}
+	if err := a.updateParentLabelsAndAnnotations(ctx, kapplyset, "latest"); err != nil {
```
justinsb, line 272: I think we only need to do this if we pruned (?)

applylib/applyset/applyset.go
```diff
@@ -153,11 +242,151 @@ func (a *ApplySet) ApplyOnce(ctx context.Context) (*ApplyResults, error) {
...
+	u, ok := obj.(*unstructured.Unstructured)
+	if !ok {
+		return fmt.Errorf("unable to convert `ApplyableObject` to `unstructured.Unstructured` %v/%v %v",
+			obj.GetName(), obj.GetNamespace(), obj.GroupVersionKind().String())
+	}
+	labels := u.GetLabels()
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+	newLabels := mergeMap(applysetLabels, labels)
```
justinsb, line 299: mergeMap is nice, but it might be clearer just to inline it (e.g. here I think you're checking if `labels == nil` twice)

applylib/applyset/applyset.go
```diff
@@ -153,11 +242,151 @@ func (a *ApplySet) ApplyOnce(ctx context.Context) (*ApplyResults, error) {
...
+		pruneObject := &pruneObjects[i]
+
+		name := pruneObject.Name
+		namespace := pruneObject.Namespace
+		mapping := pruneObject.Mapping
+		gvk := pruneObject.Object.GetObjectKind().GroupVersionKind()
+		nn := types.NamespacedName{Namespace: namespace, Name: name}
+
+		if err := a.client.Resource(mapping.Resource).Namespace(namespace).Delete(ctx, name, a.deleteOptions); err != nil {
+			klog.Error("unable to delete resource ")
```
justinsb, line 337: You might not need to log this at all, but if you do you should include `err`

applylib/applyset/applyset.go
```diff
@@ -153,11 +242,151 @@ func (a *ApplySet) ApplyOnce(ctx context.Context) (*ApplyResults, error) {
...
+	if err != nil {
+		return err
+	}
+	//kubectlapply requires the tooling and id to exist.
+	{
+		annotations := object.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+		annotations[kubectlapply.ApplySetToolingAnnotation] = a.tooling
```
justinsb, line 363: Yes, I find this much easier to read than mergeMap

applylib/applyset/applyset.go
```diff
@@ -153,11 +242,151 @@ func (a *ApplySet) ApplyOnce(ctx context.Context) (*ApplyResults, error) {
...
+	}
+	// This is not a cluster fetch. It builds up the parents labels and annotations information in kapplyset.
+	if err := kapplyset.FetchParent(a.parent.GetSubject()); err != nil {
+		return err
+	}
+
+	return a.updateParent(ctx, original, object)
+}
+
+func (a *ApplySet) updateParent(ctx context.Context, original, current metav1.Object) error {
```
justinsb, line 384: Maybe rename to updateParentLabelsAndAnnotations ?

And actually if you wanted to, you could inline the functionality of mergeMap into this function, so it would take `labels` and `annotations` as parameters

applylib/applyset/applyset.go
```diff
@@ -153,11 +242,151 @@ func (a *ApplySet) ApplyOnce(ctx context.Context) (*ApplyResults, error) {
...
+	if !reflect.DeepEqual(original.GetLabels(), current.GetLabels()) || !reflect.DeepEqual(original.GetAnnotations(), current.GetAnnotations()) {
+		if err := a.parentClient.Update(ctx, current.(client.Object)); err != nil {
+			return fmt.Errorf("error updating parent %v", err)
```
justinsb, line 387: "...: %w"

pkg/patterns/declarative/pkg/applier/applylib.go
```diff
@@ -97,3 +123,13 @@ func (a *ApplySetApplier) Apply(ctx context.Context, opt ApplierOptions) error {
 
 	return nil
 }
+
+// NewParentRef maps a declarative object's information to the ParentRef defined in the applyset library.
+func NewParentRef(restMapper meta.RESTMapper, object runtime.Object, name, namespace string) applyset.Parent {
+	gvk := object.GetObjectKind().GroupVersionKind()
+	restMapping, err := restMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+	if err != nil {
+		return nil
```
justinsb, line 132: Return an error, I think

pkg/patterns/declarative/reconciler.go
```diff
@@ -63,6 +64,8 @@ type Reconciler struct {
 
 	restMapper meta.RESTMapper
 	options    reconcilerParams
+	// preferredRestMapper caches all the RESTMappings for manifests since the Reconciler runs.
+	preferredRestMapper meta.RESTMapper
```
justinsb, line 68: Can we just use restMapper?

pkg/restmapper/caching.go
```diff
@@ -70,12 +72,35 @@ type cachedGVR struct {
 	Scope    meta.RESTScope
 }
 
+// KindFromGVR finds out the Kind from the GVR in the cache. If the GVR version is not given, we will iterate all the matching
```
justinsb, line 75: Maybe rename to KindFor, as this is a fairly direct mapping to the KindFor method on RESTMapper

pkg/restmapper/caching.go
```diff
@@ -70,12 +72,35 @@ type cachedGVR struct {
...
+func (c *cache) KindFromGVR(gvr schema.GroupVersionResource) string {
+	if gvr.Version != "" {
+		cachedgvr, ok := c.groupVersions[gvr.GroupVersion()]
+		if !ok {
+			return ""
+		}
+		return cachedgvr.toKind[gvr.Resource]
+	}
+	for keyGVR, cachedgvr := range c.groupVersions {
+		if keyGVR.Group != gvr.Group {
```
justinsb, line 87: Do we really do this in KindFor?  It seems that would often return an error, given the docs for KindFor say "KindFor takes a partial resource and returns the single match.  Returns an error if there are multiple matches"

pkg/restmapper/controllerrestmapper.go
```diff
@@ -56,7 +60,22 @@ func (m *ControllerRESTMapper) ResourcesFor(input schema.GroupVersionResource) (
 // RESTMapping identifies a preferred resource mapping for the provided group kind.
 func (m *ControllerRESTMapper) RESTMapping(gk schema.GroupKind, versions ...string) (*meta.RESTMapping, error) {
 	ctx := context.TODO()
-
+	// Since versions is optional string slice, it can be empty. If version is not given, we will iterate all the cached
+	// GV and find the first matching RESTMapping.
+	if len(versions) == 0 {
+		for keyGV := range m.cache.groupVersions {
```
justinsb, line 66: I _think_ the logic here should be identical to the logic for RESTMappings except you always return the preferred (first) version.

pkg/test/testreconciler/simpletest/testdata/reconcile/direct/create/expected-http.yaml
justinsb: Ah - this looks better now :-)

pkg/test/testreconciler/simpletest/testdata/reconcile/ssa/create/expected-http.yaml
```diff
@@ -164,27 +224,233 @@ Accept: application/json
...
+Cache-Control: no-cache, private
+Content-Length: 583
+Content-Type: application/json
+Date: (removed)
+
+{"apiVersion":"v1","data":{"k1":"v1"},"kind":"ConfigMap","metadata":{"creationTimestamp":"2022-01-01T00:00:02Z","labels":{"addons.example.org/simpletest":"simple1","applyset.kubernetes.io/part-of":"applyset-xbxAWnAItX3p1Gxrs86F-ZQAGwGoys9xxQGK3IED7bY-v1","example-app":"simpletest","l1":"v1"},"name":"foo","namespace":"ns1","ownerReferences":[{"apiVersion":"addons.example.org/v1alpha1","blockOwnerDeletion":true,"controller":true,"kind":"SimpleTest","name":"simple1","uid":"00000000-0000-0000-0000-000000000002"}],"resourceVersion":"4","uid":"00000000-0000-0000-0000-000000000003"}}
+
+---
+
+PATCH http://kube-apiserver/apis/apps/v1/namespaces/ns1/deployments/mydeployment?fieldManager=SimpleTest-applyset&force=true
```
justinsb, line 425: Looks like we're still "flapping" on this test case

applylib/applyset/applyset.go
yuwenma: done.

applylib/applyset/applyset.go
yuwenma: added. (well, it's impossible)

applylib/applyset/applyset.go
yuwenma: Sure. I was thinking that updating the manifests labels is purely for pruning purpose, pruning may not even be enabled so ignoring the errors can at least guarantee all manifests are deployed.

applylib/applyset/applyset.go
yuwenma: This should be moved to the `BeforeApply`.

applylib/applyset/applyset.go
yuwenma: Oh, that's really a good point! Right, the on-track GVRs for pruning can just be "well-prepared" but does not have to do the actual pruning work!

applylib/applyset/applyset.go
yuwenma: Sure. But the mergeMap is called 3 times, the code could be a little bit repeating.

applylib/applyset/applyset.go
yuwenma: Sure, removed.

applylib/applyset/applyset.go
yuwenma: 😄  Sure.

applylib/applyset/applyset.go
yuwenma: done. merged to updateParentLabelsAndAnnotations because this method now is only called once.

applylib/applyset/applyset.go
yuwenma: done.

pkg/patterns/declarative/pkg/applier/applylib.go
yuwenma: done. This method is used to initialize the `ApplierOptions` which does not require Parent (unless the applier is `ApplysetApplier`) . But I think you are right, even if the parent is not used, its RESTMapping should still exist.

pkg/patterns/declarative/reconciler.go
yuwenma: Done.

pkg/restmapper/controllerrestmapper.go
yuwenma: I'll revert the change for ControllerRESTMapper, since it's unrelated to this PR.

pkg/restmapper/caching.go
yuwenma: I reverted the code unrelated to this PR.

pkg/restmapper/caching.go
yuwenma: reverted.

pkg/test/testreconciler/simpletest/testdata/reconcile/ssa/create/expected-http.yaml
yuwenma: ack.

applylib/applyset/results.go
justinsb: Yes, I think we should treat pruning failures like other failures.  We might want to surface the status specially, but we shouldn't pretend everything is OK.

justinsb (approved):

justinsb:
Thanks - this looks great now, and appears to be additive, so I propose we merge and iterate on anything we find.

/approve
/lgtm

PR #326, opened 2023-04-06
Remove named return values in reconciler

justinsb:
Fixes a recent regression where we would not return errors from reconcileExists.


yuwenma:
/lgtm 

The PR is great!

PR #328, opened 2023-04-10
restmapper: Ensure that we only access cached fields from a single method

justinsb:
Also ensure this access is always guarded with a mutex.

### DIFF
--- a/pkg/restmapper/caching.go
+++ b/pkg/restmapper/caching.go
@@ -17,32 +17,32 @@ import (
 
 // cache is our cache of schema information.
 type cache struct {
-	mutex         sync.Mutex
-	groups        map[string]metav1.APIGroup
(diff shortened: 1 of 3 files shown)


pkg/restmapper/caching.go
```diff
@@ -17,124 +17,157 @@ import (
...
 	}
-	return &meta.RESTMapping{
-		Resource:         c.gv.WithResource(cached.Resource),
-		GroupVersionKind: c.gv.WithKind(kind),
-		Scope:            cached.Scope,
-	}, nil
+	return nil, nil
 }
 
 // fetch returns the metadata, fetching it if not cached.
```
atoato88, line 162:
```suggestion
// fetchServerResources returns the metadata, fetching it if not cached.
```
This method is renamed to `fetchServerResources`.
But it's small point for implementing.

PR #330, opened 2023-04-21
fix: bump applylib in go mod

yuwenma:
This is critical for users who use latest k-d-p.


justinsb:
Thanks

/approve
/lgtm

PR #331, opened 2023-05-03
feat: avoid importing kubectl deps when using ApplysetApplier

yuwenma:
This PR allows k-d-p operators to not import the kubectl dependent libraries if using `ApplysetApplier`. 

**CI change and test coverage**:
Some Test files are tagged with `without_kubectl_deps` specifically. We added a new CI test run in the `dev/test` to make sure those tests are run with the  `without_kubectl_deps` tag 
```
CGO_ENABLED=0 go test -tags without_kubectl_deps -count=1 -v ./...
``` 

**usage**:

Take Guestbook as an example, 
```
// examples/guestbook-operator/controllers/guestbook_controller.go
applier := applier.NewApplySetApplier(metav1.PatchOptions{}, metav1.DeleteOptions{}, applier.ApplysetOptions{})
return r.Reconciler.Init(mgr, &api.Guestbook{},
        ...
        declarative.WithApplier(applier),
		declarative.WithApplyPrune())
```

`go run -tags without_kubectl_deps main.go`


pkg/patterns/declarative/pkg/applier/golden_test.go
```diff
@@ -0,0 +1,155 @@
...
+func fakeParent() runtime.Object {
+	parent := &unstructured.Unstructured{}
+	parent.SetKind("ConfigMap")
+	parent.SetAPIVersion("v1")
+	parent.SetName("test")
+	parent.SetNamespace("default")
+	return parent
+}
+
+func runApplierGoldenTests(t *testing.T, testDir string, interceptHTTPServer bool, applier Applier) {
```
yuwenma, line 35: All 3 appliers needs this method. So I move it from `applyset.go` to non-tagged file

pkg/test/testreconciler/simpletest/controller_test.go
```diff
@@ -28,33 +25,6 @@ import (
 	api "sigs.k8s.io/kubebuilder-declarative-pattern/pkg/test/testreconciler/simpletest/v1alpha1"
 )
 
-func TestSimpleReconciler(t *testing.T) {
```
yuwenma, line 31: split the two test cases in `TestSimpleReconciler` to `with_kubectl_test.go` and `without_kubectl_test.go` so that we can use different tags for each file.

pkg/patterns/declarative/pkg/applier/direct_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build !without_kubectl_deps
```
justinsb, line 1: Maybe we do this by applier, because there are two "dependencies" on kubectl - there is execing kubectl, and there is calling into the kubectl code.

So maybe we have two tags: `without_kubectl_applier` (or `without_exec_applier`) and `without_direct_applier` ?

pkg/patterns/declarative/pkg/applier/golden_test.go
justinsb: OK, but it might be easier just not to tag that file, if that's an option.

pkg/test/testreconciler/simpletest/controller_test.go
justinsb: sgtm!

justinsb (commented):
Looks good, but I think if we can avoid `go:build without_kubectl_deps` except in one file that builds the default object, I think this PR gets simpler.

yuwenma:
> Looks good, but I think if we can avoid `go:build without_kubectl_deps` except in one file that builds the default object, I think this PR gets simpler.

That's true. My concern is that if we do not use `!without_kubectl_deps`, the CI run with and without tags will both run the kubectl direct and exec use cases.

pkg/patterns/declarative/pkg/applier/applylib.go
```diff
@@ -123,13 +125,3 @@ func (a *ApplySetApplier) Apply(ctx context.Context, opt ApplierOptions) error {
 
 	return nil
 }
-
-// NewParentRef maps a declarative object's information to the ParentRef defined in the applyset library.
-func NewParentRef(restMapper meta.RESTMapper, object runtime.Object, name, namespace string) (applyset.Parent, error) {
```
yuwenma, line 128: moved to `global_without_kubectl.go`

yuwenma:
> Looks good, but I think if we can avoid `go:build without_kubectl_deps` except in one file that builds the default object, I think this PR gets simpler.

qq: but I think this will cause the `without_kubectl_deps` test to call the test cases that need the kubectl dep (which cause undefined error because the direct.go and  exec.go are excluded).
e.g. 
run `go test -tags without_direct_applier,without_exec_applier` runs
``` with_kubectl_test.go 
// assuming I commented out the tags
//go:build !without_exec_applier || !without_direct_applier
// +build !without_exec_applier !without_direct_applier

func TestSimpleReconciler(t *testing.T) {
	Key := "direct"
	t.Run(Key, func(t *testing.T) {
		testharness.RunGoldenTests(t, "testdata/reconcile/"+Key+"/", func(h *testharness.Harness, testdir string) {
            // ! NewDirectApplier raises undeclared error
			testSimpleReconciler(h, testdir, applier.NewDirectApplier(), status.NewBasic(nil)) 
		})
	})
}
```

pkg/patterns/declarative/metrics_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build without_kubectl_deps
+// +build without_kubectl_deps
```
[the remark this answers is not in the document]
justinsb: The Applyset applier should always be present, I think.

pkg/patterns/declarative/pkg/applier/applylib_test.go
```diff
@@ -1,162 +1,16 @@
+//go:build without_exec_applier && without_direct_applier
```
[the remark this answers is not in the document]
yuwenma: reverted the change in applylib.go and applylib_test.go

justinsb (approved):

justinsb:
Thanks @yuwenma - lgtm!

/approve
/lgtm

PR #342, opened 2023-06-26
Create abstraction layer over controller-runtime/client-go versions

justinsb:
Create an abstraction layer so that we are less dependent on a particular version of client-go / controller-runtime.  The intent here is that controllers can work with multiple versions of client-go/controller-runtime, and thus a wide range of k8s versions.

- Make our applyset not depend on generics
- Create commonclient to abstract controller-runtime versions


yuwenma:
Thank you so much for adding this abstraction layer. LGTM (looks like the test caught some additional cases need to fix).

PR #350, opened 2023-09-05
WIP: ✨ A yaml-based DSL for creating CRDs

justinsb:
Aiming to make it easier to create CRDs.


yamlcrd/pkg/parser/testdata/simple1/expected.yaml
```diff
@@ -0,0 +1,70 @@
+XNamespace:
```
apelisse, line 1: it's hard to compare input with output since none of your fields appear to be omitempty in the output.

yamlcrd/pkg/parser/testdata/simple1/input.yaml
```diff
@@ -0,0 +1,22 @@
+XNamespace:
```
apelisse, line 1: is this example using all the features you've implemented? I'm curious how some of the complicated corner-cases are going to turn out?

yamlcrd/pkg/parser/testdata/simple1/input.yaml
```diff
@@ -0,0 +1,22 @@
...
+    # instanceType is an instanceType
+    instanceType: string
+
+    # minReplicas is the minimum number of replicas
+    minReplicas: int(min=0)
+
+    # maxReplicas is the maximum number of replicas
+    maxReplicas: int(min=0)
+
+    containers:
```
apelisse, line 15: For example, that list should have a listMap or whatever, right?

yamlcrd/pkg/parser/testdata/simple1/input.yaml
```diff
@@ -0,0 +1,22 @@
...
+    labels: map(key=string,value=string)
+
+    # instanceType is an instanceType
+    instanceType: string
+
+    # minReplicas is the minimum number of replicas
+    minReplicas: int(min=0)
+
+    # maxReplicas is the maximum number of replicas
+    maxReplicas: int(min=0)
```
apelisse, line 13: I'm assuming the `int(min=0)` language is entirely made-up? What are some of the principles behind it?

yamlcrd/pkg/parser/testdata/simple1/expected.yaml
justinsb: Good point!  And fixed - I was importing the wrong yaml library (yaml.v3 instead of sigs.k8s.io/yaml)

yamlcrd/pkg/parser/testdata/simple1/input.yaml
justinsb: So there's definitely some major cheating going on right now, things like `map(key=string,value=string)` and `int(min=0)` are hard-coded so we can iterate on the syntax before committing to it.

Are there any particular corner scenarios you have in mind?

yamlcrd/pkg/parser/testdata/simple1/input.yaml
justinsb: Inspired by your kabstract python syntax.  I think that named parameters are much better than positional ones here.  We don't have anything particularly relevant in yaml.  I guess that maybe we could do something like `int{ min: 0 }` but I think that might actually interact relatively poorly with yaml - I'm trying to do things that aren't too painful with yaml escaping rules etc.

I guess one big question is what about string values, how will quoting work there - I guess I'll try adding a string or two :-)

justinsb:
Wrote a (slightly) better parsing function for arguments, and also implemented a few more trickier annotations:

```
    containers(map-key=name):
    - name: string # The identifier for this container
      image: string(regex="image.*") # The container image to run
```

I don't love the placement of the map-key, but it does show that we can put parameters on keys without any real problem from the yaml parser.

PR #352, opened 2023-09-11
:warning: Bump controller-runtime to 0.16.1

justinsb:
- Update controller-runtime to 0.16.1
- Update code for controller-runtime 0.16.0
- Deprecate client-side validation in the direct applier

:warning: Client-side validation is now deprecated in the direct applier.  Server-side validation should be used instead, please open an issue if you have a use-case for client-side validation.


.github/workflows/main.yml
```diff
@@ -47,7 +47,7 @@ jobs:
 
       - uses: actions/checkout@v2
 
-      - run: go mod tidy
+      - run: dev/format-gomod
```
yuwenma, line 50: that's nit.

yuwenma (approved):
Looks great! Thank you for keeping this  up-to-date with  controller-runtime.

PR #353, opened 2023-09-21
Play nicely with third-party testing libraries

tomasaschan:

This makes the validator use an `interface` compatible with `*testing.T` instead of using `*testing.T` directly, in order to allow third-party testing libraries to provide an equivalent implementation instead of the built-in one.

See e.g. https://onsi.github.io/ginkgo/#using-third-party-libraries for a use case of this setup.


tomasaschan:
Ping @justinsb :)

atoato88:
/lgtm
It looks good, and pased the test.

tomasaschan:
Ping! Could someone please /approve this? Would be nice to get it into `master` so we can use this in our operator test suites.

PR #355, opened 2023-09-24
mockkubeapiserver: Refactor storage to be pluggable

justinsb:
This helps if we want e.g. to capture all the objects easily.


yuwenma (approved):
Since we found a data race in the previous memorystorage,  can we try `go test -race` and validate the new change passes the data race? (I think so because we keep the storage mutex in each CRUD method)

mockkubeapiserver/storage/interface.go
```diff
@@ -0,0 +1,33 @@
+package storage
```
yuwenma, line 1: I like this style in storage, and the hook (I misread the rationale of Hook until seeing the "CRDHook")

mockkubeapiserver/storage/memorystorage/crd.go
```diff
@@ -0,0 +1,101 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/kubebuilder-declarative-pattern/mockkubeapiserver/storage"
+)
+
+// UpdateCRD is called whenever a CRD is updated.
+// We register the types from the CRD.
+func (s *MemoryStorage) UpdateCRD(ev *storage.WatchEvent) error {
+	// TODO: Deleted / changed CRDs
```
yuwenma, line 16: What is the expected behavior of "delete CRD". Is this for the case when CRD belongs to a versioned manifest?

mockkubeapiserver/storage/memorystorage/memorystorage_watch.go
```diff
@@ -0,0 +1,119 @@
...
+	mutex   sync.Mutex
+	watches []*watch
+
+	resourceVersionClock *resourceVersionClock
+
+	objects map[types.NamespacedName]*unstructured.Unstructured
+}
+
+type resourceVersionClock struct {
+	now atomic.Int64
```
yuwenma, line 33: I missed the atomic in the previous review. This is cool.

mockkubeapiserver/storage/memorystorage/memorystorage_watch.go
```diff
@@ -0,0 +1,119 @@
...
+
+	r.mutex.Lock()
+	for _, obj := range r.objects {
+		if opt.Namespace != "" {
+			if obj.GetNamespace() != opt.Namespace {
+				continue
+			}
+		}
+
+		ev := storage.BuildWatchEvent(gvk, "ADDED", obj)
```
yuwenma, line 84: Just to make sure I understand the code correctly: this "ADDED" is a no-op and here we mainly use the watchCallback (next line), right?

mockkubeapiserver/hooks/deployment.go
```diff
@@ -0,0 +1,98 @@
...
+	if statusObj == nil {
+		statusObj = make(map[string]interface{})
+		u.Object["status"] = statusObj
+	}
+	status, ok := statusObj.(map[string]interface{})
+	if !ok {
+		return fmt.Errorf("status was of unexpected type %T", statusObj)
+	}
+
+	generation := u.GetGeneration()
```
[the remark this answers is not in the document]
justinsb: Good question :-)  I think I added it because otherwise kstatus will not consider the deployment ready (which makes our golden tests unhappy / means that they don't converge)

mockkubeapiserver/hooks/namespace.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Fake the required transitions for Namespace objects so they become ready
+func (s *NamespaceHook) namespaceChanged(ev *storage.WatchEvent) {
+	u := ev.Unstructured()
+
+	// These changes seem to be done synchronously (similar to a mutating webhook)
+	labels := u.GetLabels()
+	name := u.GetName()
+	if labels["kubernetes.io/metadata.name"] != name {
```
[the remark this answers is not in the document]
justinsb: Yeah, I think it is legacy, but ... it's there.  I think we prefer field selectors now: https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/

Only a handful of fieldSelectors are implemented for some objects (though I think it would be cool if we supported arbitrary CEL expressions).  fieldSelector for `metadata.name` is implemented everywhere I believe, because that is how we can watch one object (you can only pass the `watch` parameter to LIST requests, IIRC)

mockkubeapiserver/storage/memorystorage/crd.go
justinsb: If a CRD is deleted, we should unregister it from the apiserver, so it isn't discoverable, you can't query it etc.  That's ... not implemented yet!

mockkubeapiserver/storage/memorystorage/memorystorage_watch.go
justinsb: Yes, in general it's premature optimization vs sync.Mutex, but here I find it more ergonomic (as well as being possibly more efficient)

mockkubeapiserver/storage/memorystorage/memorystorage_watch.go
justinsb: This is the "real" watch code, the hooks are a sort of "fast-path built-in watch".  When you start a watch with kube-apiserver with resourceVersion unspecified, you get a synthetic ADDED event for every object that exists at that time.

It's actually a streaming LIST (which is otherwise missing from kube), and a little more efficient on the client (at least) in terms of memory, because you only have to process one object at a time, instead of a whole page of objects.  The downside is that it's hard to tell when the initial LIST is over, and you have all the objects.  In the past few versions of kube-apiserver, they added a bookmark notification when the initial list was completed, which is how you can tell.  Maybe I should add bookmark notifications :-)  More info here: https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/3157-watch-list/README.md

mockkubeapiserver/hooks/deployment.go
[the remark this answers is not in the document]
yuwenma: Ah, makes sense.

PR #357, opened 2023-10-23
fix: remove client-side validation from example guestbook operator

ChristopherFry:
This pull request removes the client-side apply validation from the example guestbook operator allowing the operator to successfully reconcile changes.

Previously the operator would show this error when reconciling:
```
2023-10-22T21:39:20-07:00       ERROR   applying manifest       {"controller": "guestbook-controller", "object": {"name":"guestbook-sample","namespace":"default"}, "namespace": "default", "name": "guestbook-sample", "reconcileID": "1063a260-c92e-43ea-b82f-cf2114cdbeef", "error": "client-side validation is no longer supported"}
```

This error is due to client-side validation being deprecated in the direct applier with #352.

### DIFF
--- a/examples/guestbook-operator/controllers/guestbook_controller.go
+++ b/examples/guestbook-operator/controllers/guestbook_controller.go
@@ -62,7 +62,6 @@ func (r *GuestbookReconciler) setupReconciler(mgr ctrl.Manager) error {
 		declarative.WithApplyKustomize(),
 
 		// Add other optional options for testing
-		declarative.WithApplyValidation(),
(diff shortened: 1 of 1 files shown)


justinsb:
Thanks @ChristopherFry!

This reminds me of an idea that we've had but never written down, IIRC: we should log a warning when users use things/don't use things we recommend.  Here it's pretty clear-cut and gives an obvious error, but it would still be better if that warning happened earlier.  With an env-var, we could even switch it between warning or fatal or silent, corresponding to development vs production.

In any case...

/approve
/lgtm

PR #358, opened 2023-10-23
chore: allow tools to be installed with newer versions of go

ChristopherFry:
This change fixes the example guestbook operator Makefile by allowing the `controller-gen` and `kustomize` tools to be successfully installed.

Previously any Make commands requiring one of these tools such as `make generate` would fail since the `go-get-tool` function would only download the package with go get however not install the package with go install, where go get no longer installs executables since Go 1.17, https://go.dev/doc/go-get-install-deprecation.

The updated Makefile commands for `controller-gen` and `kustomize` aligns with the Makefile kubebuilder generates for new projects, https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v3/Makefile.

### DIFF
--- a/examples/guestbook-operator/Makefile
+++ b/examples/guestbook-operator/Makefile
@@ -87,25 +87,31 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/default | kubectl delete -f -
 
+## Location to install dependencies to
+LOCALBIN ?= $(shell pwd)/bin
+$(LOCALBIN):
+	mkdir -p $(LOCALBIN)
 
-CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
-controller-gen: ## Download controller-gen locally if necessary.
-	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1)
+## Tool Binaries
+KUSTOMIZE ?= $(LOCALBIN)/kustomize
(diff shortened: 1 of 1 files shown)


examples/guestbook-operator/Makefile
```diff
@@ -87,25 +87,31 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
...
+			echo "$(LOCALBIN)/kustomize version is not expected $(KUSTOMIZE_VERSION). Removing it before installing."; \
+			rm -rf $(LOCALBIN)/kustomize; \
+	fi
+	test -s $(LOCALBIN)/kustomize || { curl -Ss $(KUSTOMIZE_INSTALL_SCRIPT) --output install_kustomize.sh && bash install_kustomize.sh $(subst v,,$(KUSTOMIZE_VERSION)) $(LOCALBIN); rm install_kustomize.sh; }
+
+.PHONY: controller-gen
+controller-gen: $(CONTROLLER_GEN) ## Download controller-gen locally if necessary. If wrong version is installed, it will be overwritten.
+$(CONTROLLER_GEN): $(LOCALBIN)
+	test -s $(LOCALBIN)/controller-gen && $(LOCALBIN)/controller-gen --version | grep -q $(CONTROLLER_TOOLS_VERSION) || \
+	GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_TOOLS_VERSION)
```
justinsb, line 117: I think we can just `go run` these now, not sure if that's better though!

justinsb:
Thanks @ChristopherFry .  I've started using `go run` everywhere rather than trying to install things, there used to be a big performance penalty but I _think_ that penalty is now negligible.

But this simplifies the build scripts, so...

/approve
/lgtm

PR #361, opened 2023-11-21
Fix Validator output spacing #360

### LINKED ISSUE #360: Validator inserts extra newlines between Objects
**What happened**:

[Validator](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/f6a5c89244f2221088444042166f89743d409515/pkg/test/golden/validator.go#L106) expects empty lines before and after each document separator `---` in the expected output manifests that I specify in "*.out.yaml".

```
#example.out.yaml
apiVersion: v1
kind: Namespace
metadata:
  name: x

---

apiVersion: v1
kind: Namespace
metadata:
  name: y

```

I expected that I should be able to specify consecutive objects in my "*.out.yaml" file without the extra newlines.

```
#example.out.yaml
apiVersion: v1
kind: Namespace
metadata:
  name: x
---
apiVersion: v1
kind: Namespace
metadata:
  name: y

```

Standard test code that creates a `Validator` to validate a user-defined `Reconciler` should be sufficient.

```
package _

import (
	"testing"

	"sigs.k8s.io/kubebuilder-declarative-pattern/pkg/test/golden"
)

func TestXReconciler(t *testing.T) {
	validator := golden.NewValidator(t, xSchemeBuilder)
	x := XReconciler{}
	if err := x.SetupWithManager(validator.Manager()); err != nil {
		...
	}
	validator.Validate(x)
}
```

This line 

* https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/f6a5c89244f2221088444042166f89743d409515/pkg/test/golden/validator.go#L324

that adds additional newlines, even though [yamilzer.Encode](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/f6a5c89244f2221088444042166f89743d409515/pkg/test/golden/validator.go#L327) already adds a newline after the object, is likely the culprit.

annasong20:
Fixes #360 by removing extra newlines between objects in test/validator manifest output.

Fixes #360.


yuwenma (commented):
Looks good. Thanks!

atoato88:
Nice catch!
Thank you for posting PR!

PR #367, opened 2024-01-08
Allow dynamic client to be set

justinsb:
This allows us to override the dynamic client, useful for tests or if
we want to intercept and track all the calls (for automatic dependency
tracking).

### DIFF
--- a/pkg/patterns/declarative/pkg/applier/applylib.go
+++ b/pkg/patterns/declarative/pkg/applier/applylib.go
@@ -54,9 +54,13 @@ func (a *ApplySetApplier) Apply(ctx context.Context, opt ApplierOptions) error {
 
 	patchOptions.Force = &opt.Force
 
-	dynamicClient, err := dynamic.NewForConfig(opt.RESTConfig)
-	if err != nil {
-		return fmt.Errorf("error building dynamic client: %w", err)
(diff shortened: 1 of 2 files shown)


pkg/patterns/declarative/pkg/applier/applylib.go
```diff
@@ -54,9 +54,13 @@ func (a *ApplySetApplier) Apply(ctx context.Context, opt ApplierOptions) error {
...
-	dynamicClient, err := dynamic.NewForConfig(opt.RESTConfig)
-	if err != nil {
-		return fmt.Errorf("error building dynamic client: %w", err)
+	dynamicClient := opt.DynamicClient
+	if dynamicClient == nil {
+		d, err := dynamic.NewForConfig(opt.RESTConfig)
+		if err != nil {
+			return fmt.Errorf("error building dynamic client: %w", err)
+		}
+		dynamicClient = d
```
cheftako, line 63: Any reason not to set opt.DynamicClient to d in this case? Could save us need to repeatedly generating dynamicClients during repeated reconciles.

justinsb: Good question, I'll take a look!  I don't _think_ dynamic client caches any server-generated state, so it's "just memory", but I'll chekc!

PR #368, opened 2024-01-08
Extract GVK more robustly when building applyset parent

justinsb:
Include the case when we're using a typed object.

### DIFF
--- a/pkg/patterns/declarative/reconciler.go
+++ b/pkg/patterns/declarative/reconciler.go
@@ -370,9 +370,13 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 		}
 	}
 
-	parentRef, err := applier.NewParentRef(r.restMapper, instance, instance.GetName(), instance.GetNamespace())
+	gvk, err := apiutil.GVKForObject(instance, r.mgr.GetScheme())
 	if err != nil {
-		return statusInfo, err
+		return statusInfo, fmt.Errorf("getting GVK for %T: %w", instance, err)
+	}
+	parentRef, err := applier.NewParentRef(r.restMapper, instance, gvk, instance.GetName(), instance.GetNamespace())
+	if err != nil {
+		return statusInfo, fmt.Errorf("building applyset parent: %w", err)
 	}
 	applierOpt := applier.ApplierOptions{
 		RESTConfig:        r.config,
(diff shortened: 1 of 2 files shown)


maqiuyujoyce:
Thank you so much for the quick fix, @justinsb ! I made a local copy of your change and verified the issue in the CC reconciliation is fixed.

/lgtm

pkg/patterns/declarative/reconciler.go
```diff
@@ -370,9 +370,13 @@ func (r *Reconciler) reconcileExists(ctx context.Context, name types.NamespacedN
 		}
 	}
 
-	parentRef, err := applier.NewParentRef(r.restMapper, instance, instance.GetName(), instance.GetNamespace())
+	gvk, err := apiutil.GVKForObject(instance, r.mgr.GetScheme())
 	if err != nil {
-		return statusInfo, err
+		return nil, fmt.Errorf("getting GVK for %T: %w", instance, err)
```
cheftako, line 375: So this appears to be us returning an error we were never returning (ignoring) before. I believe there is also a difference between `object.GetObjectKind().GroupVersionKind()` and `apiutil.GVKForObject(instance, r.mgr.GetScheme())` such that we things which were not errors for the former (eg 0 length kind or versions) which are errors for the latter.

justinsb: client.Client uses apiutil.GVKForObject (IIRC), so we really should have been using that all along.  We got away with it up until now because the cached client seems to set the GVK in the object.  But when the KCC operator turned off the cache, the GVK was no longer set (possibly a bug in controller-runtime), and the difference mattered.

We also blew up immediately in the existing code if the GVK was empty.  The first thing NewParentRef does is try to map the GVK to a resource, that gave an error which we weren't ignoring.  By returning the error with the verbose message here, we hopefully make it easier to understand what's going wrong (and make it less likely that something is going wrong anyway, as we probably don't make it to this point if we can't make the object to a GVK!)

I did fix from returning `nil, fmt.Errorf...` to returning `statusInfo, fmt.Errorf...` ; mostly for consistency but it is more correct also.

maqiuyujoyce: Just to add some data points: During the upgrade of KDP from v0.13.0-beta to v0.15.0-beta, KCC had to deal with 12+/20+ (depending on how they are counted) breaking changes. Not saying that it is a good thing to have breaking changes, though if it is a fix and is potentially breaking, I think it seems to be aligning with the philosophy of KDP's minor version bump.

justinsb: @maqiuyujoyce I think the majority of those breakages are due to controller-runtime; we are actually starting to build a compatibility layer into kdp to try to smooth these transitions (`commonclient`).  This one though is a bug/regression, although we are also switching the apply method which is why we hit it (and that should be opt-in).

PR #371, opened 2024-01-16
Optimization: Reuse HTTPClient / DynamicClient in reconcile

justinsb:
Should be slightly more efficient via better HTTP connection reuse.

### DIFF
--- a/commonclient/factory_cr11.go
+++ b/commonclient/factory_cr11.go
@@ -4,10 +4,13 @@ package commonclient
 
 import (
 	"context"
+	"net/http"
 
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/util/workqueue"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/cluster"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/source"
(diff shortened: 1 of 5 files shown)


justinsb:
cc @cheftako  - as requested re reusing the DynamicClient; I think the answer is that the caller should try to provide a cached DynamicClient, because client-go is more efficient if we can reuse the `http.Client`.  Added a comment and ensured that k-d-p itself passes the dynamic client (which we actually already have!)

atoato88:
I am not sure about my comment is related with this PR. :thinking: 

Current direct applier uses the `dynamicClient` at [here](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/8e6cf52b8fdd8dc1bb1516529c994ccb46cd9ab7/pkg/patterns/declarative/pkg/applier/direct.go#L92).
This always uses the dynamicClient from Factory.
Should we update the `direct.go` to use the `opt.DynamicClient`?

On the other hand, [`dynamicClient` in `applylib.go`](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/8e6cf52b8fdd8dc1bb1516529c994ccb46cd9ab7/pkg/patterns/declarative/pkg/applier/applylib.go#L58-L65) takes care of the existance of `opt.DynamicClient`.
It's OK as-is, I think.

justinsb:
Great point @atoato88, and you're absolutely right.  I added a commit that should also reuse the dynamic-client in the direct applier.

(I don't think we can practically reuse it in the exec applier.)

I also noticed that we probably should be trying to reuse the http client [here](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/8e6cf52b8fdd8dc1bb1516529c994ccb46cd9ab7/pkg/patterns/declarative/watch.go#L114), but I think I'll have to send a separate PR for that one!

atoato88:
@justinsb 
Thank you for the response and update.

> (I don't think we can practically reuse it in the exec applier.)

I think so too.

> I also noticed that we probably should be trying to reuse the http client [here](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/8e6cf52b8fdd8dc1bb1516529c994ccb46cd9ab7/pkg/patterns/declarative/watch.go#L114), but I think I'll have to send a separate PR for that one!

OK!
We can go forward. :rocket:

PR #373, opened 2024-01-24
Switch to after apply when setting up watches

barney-s:
This change is to workaround a potential bug where we are setting up watches in a namespace that is not yet created by apply.

Fixes #

**Additional documentation**:


barney-s:
We are root causing the issue. Both Before and After apply should work

PR #376, opened 2024-01-27
Close the WatchStart before starting the watch. It was anyways closed…

barney-s:
Setting up Watch() sometimes results in Watch() being a blocking call. Add a cancellable context that times out sooner when there is no activity in the Apply() path. Otherwise it timesout in 5mins. Cancellable context is better than watch options.Timeout since watch timeout happens even if there is activity on the watch channel. 

In BeforeApply watch setup case, it blocks the Apply loop if the API server doesnt have any watch events for the first object being applied.

Fixes #

**Additional documentation**:


pkg/patterns/declarative/pkg/watch/dynamic.go
```diff
@@ -35,8 +36,17 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-// WatchDelay is the time between a Watch being dropped and attempting to resume it
-const WatchDelay = 30 * time.Second
+const (
```
atoato88, line 39: There are two `const` sections at here.
What is the reason to be splited to?

atoato88:
Current title of this PR doesn't indicate the current contents, I think.
Could you update the PR title?

PR #378, opened 2024-02-05
WIP: Add developer hints

justinsb:
WIP and distinct from #377 as I imagine we will want more discussion e.g. about testing / how to turn it off/on etc.

We've evolved a lot over time, and we could start nudging developers
to use the now-recommended approaches and avoid the less-recommended
approaches (without breaking compatability).
    
Introduce a "hints" package that enables us to output messages to developers to nudge them towards best practices.

(Builds on #377 as this is what made me think of this!)

### DIFF
--- a/pkg/hints/log.go
+++ b/pkg/hints/log.go
@@ -1,0 +1,33 @@
+package hints
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"k8s.io/klog/v2"
+)
+
+// EnvRecommendationsAreStrict is the environment variable that can be set
+// to make developer recommendations a panic.  It can be set by developers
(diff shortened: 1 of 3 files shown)


tomasaschan:
@justinsb a less hacky way to accomplish this would be to build a linter that can emit recommendations when obsolete methods are used - or to find one that already supports deprecation workflows and start annotating (or whatever mechanism that uses) APIs that now have better alternatives to support that. Then any developer using something like golangci-lint gets the hints without having to introduce new workflows or runtime panics...

justinsb:
Hmm, I missed your suggestion @tomasaschan  - reopening so I can try that.  IIRC some of the things we probably want to detect are hard to detect statically.

PR #379, opened 2024-02-07
Bump k8s.io/kubectl to 0.29.1

tomasaschan:
Main motivation is to include https://github.com/kubernetes/kubernetes/pull/122346

I don't know if there's a way to do this that touches fewer dependencies; I tried a bunch of ways with `go get [-u] k8s.io/kubectl[@0.29.1]` but all of them resulted in this full set of changes.

PR #380, opened 2024-02-26
Bump controller-runtime to 0.17 / kube to 1.29

justinsb:
- chore: Update to controller-runtime 0.17 / kube 1.29
- deprecation: Use NewDynamicRESTMapper instead of NewDiscoveryRESTMapper


tomasaschan:
Let me know if I can help speed this along somehow!

atoato88:
This PR is related with the content of #379.
I think #379 can be closed if this PR is merged.

go.mod
```diff
@@ -1,6 +1,8 @@
 module sigs.k8s.io/kubebuilder-declarative-pattern
 
-go 1.20
+go 1.21
+
+toolchain go1.22.0
```
atoato88, line 5: Maybe, we should update the files for GitHub Actions?
- https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/.github/workflows/main.yml
This file use the go v1.20 setup at 2 place.

atoato88 (commented):
I know this PR is WIP, but I left just one comment.

atoato88:
@justinsb 
Ping, how about the current state of this?

go.mod
justinsb: You're right - I just pushed that and tests are looking much happier!

justinsb: And I just saw that @tomasaschan proposed a better version of the change in https://github.com/justinsb/kubebuilder-declarative-pattern/pull/1 , so using that one

justinsb:
Removed WIP, and I also think we should consider expanding ownership of this repo, perhaps starting with @tomasaschan .  Sometimes I get pulled into different things and things fall off my plate - sorry for the delays here!

tomasaschan:
/lgtm

(but I suspect it won't matter yet...)

PR #382, opened 2024-04-02
Have the ability to set Watch timeouts

barney-s:
Add the ability to set watch timeouts.

Allow setting shorter watch timeout for the firt time alone.

 We have seen in some evnironments (where requests are tunnelled) setting up the (first) watch is blocking if there are no resources available to watch. This causes the apply loop to be blocked when setting up watches before apply.


haiyanmeng:
@barney-s , why do we no longer need this change?

PR #383, opened 2024-04-17
mockkubeapiserver: Support stringData when creating a secret

justinsb:
This is an edge case in the kube apiserver, but there is special
handling for the stringData field of a secret, that is mapped to
base64 data.


justinsb:
Hmmm .... this no longer needs a rebase (at least it shouldn't)...  I'll try a retest

/retest

mockkubeapiserver/postresource.go
```diff
@@ -73,8 +74,42 @@ func (req *postResource) Run(ctx context.Context, s *MockKubeAPIServer) error {
...
 }
+
+func preCreateHook(ctx context.Context, obj *unstructured.Unstructured) error {
+	gvk := obj.GroupVersionKind()
+	if gvk.Group == "" && gvk.Version == "v1" && gvk.Kind == "Secret" {
+		stringDataAny, found := obj.Object["stringData"]
+		if found {
+			stringDataMap, ok := stringDataAny.(map[string]any)
+			if !ok {
+				return fmt.Errorf("unexpected type for v1.Secret stringData, got %T", stringDataAny)
```
tomasaschan, line 94:
```suggestion
				return fmt.Errorf("unexpected type for v1.Secret stringData: expected map[string]any, got %T", stringDataAny)
```

mockkubeapiserver/postresource.go
```diff
@@ -73,8 +74,42 @@ func (req *postResource) Run(ctx context.Context, s *MockKubeAPIServer) error {
...
+			}
+			dataAny, dataFound := obj.Object["data"]
+			if !dataFound {
+				dataAny = make(map[string]any)
+				obj.Object["data"] = dataAny
+			}
+			dataMap, ok := dataAny.(map[string]any)
+			if !ok {
+				return fmt.Errorf("unexpected type for v1.Secret data, got %T", dataAny)
```
tomasaschan, line 103:
```suggestion
				return fmt.Errorf("unexpected type for v1.Secret data: expected map[string]any, got %T", dataAny)
```

mockkubeapiserver/postresource.go
```diff
@@ -73,8 +74,42 @@ func (req *postResource) Run(ctx context.Context, s *MockKubeAPIServer) error {
...
+			}
+			for k, vAny := range stringDataMap {
+				v, ok := vAny.(string)
+				if !ok {
+					return fmt.Errorf("unexpected type for v1.Secret stringData entry, got %T", vAny)
```
tomasaschan, line 108:
```suggestion
					return fmt.Errorf("unexpected type for v1.Secret stringData entry %s: expected string, got %T", k, vAny)
```

mockkubeapiserver/postresource.go
```diff
@@ -73,8 +74,42 @@ func (req *postResource) Run(ctx context.Context, s *MockKubeAPIServer) error {
...
+				}
+				dataMap[k] = base64.StdEncoding.EncodeToString([]byte(v))
```
tomasaschan, line 110: I'm not familiar enough with what Go will do through all of these casts; will this actually end up mutating `obj`? Or does it just mutate the `dataMap` we've created, but not actually propagate those changes back?

tomasaschan (commented):
A bit of a nit, but still: the errors could be more descriptive.

Also, wouldn't hurt to cover/document this behavior in tests :)

justinsb:
> Also, wouldn't hurt to cover/document this behavior in tests :)

A great point, I was thinking about that the other day.  This change was to support testing in k8s-config-connector, so it got test coverage there, but we could easily create some simple tests directly in the repo - let me do that!

mockkubeapiserver/postresource.go
justinsb: Thanks, agree 100%

mockkubeapiserver/postresource.go
justinsb: So the answer is yes, and the test will (hopefully!) prove it.  I'll have a go at rewriting it to be more self-evidently true though, I agree there's too much happening here.

justinsb:
This is escalating a bit ... looks like I need to break out a new module to avoid a circular dependency: https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/pull/398

justinsb:
Had a bit of trouble with the go.sum / go.mod checking (dev/format-gomod), but I finally figured out that the go.sum difference was to do with my cache; I did `go clean -cache -modcache` before running `dev/format-gomod`.  I think it's still worth trying go.work (and I added a note to the dev/format-gomod script and to the GitHub action error message)

justinsb:
This _might_ be much easier if we merged a (basically empty) `ktest` module first - then I think we could stay away from `go work sync` which is the root of all the problems.  Maybe that's a reasonable compromise that's worth a go.

justinsb:
I tried staging the ktest module in #400.  I think that would mean I could back out the `go work sync` logic which gave me so much trouble

justinsb:
Found a repo that checks in go.work: kubernetes/kubernetes !  I think we're on solid ground :-)

ktest/testharness/manifest.go
```diff
@@ -0,0 +1,52 @@
...
+	"context"
+	"fmt"
+	"io"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	k8syaml "k8s.io/apimachinery/pkg/util/yaml"
+)
+
+func ParseObjects(ctx context.Context, manifest string) ([]*unstructured.Unstructured, error) {
```
yuwenma, line 30: 2 years passed and we still have to write this. 🤣

mockkubeapiserver/postresource.go
```diff
@@ -73,8 +74,55 @@ func (req *postResource) Run(ctx context.Context, s *MockKubeAPIServer) error {
...
+func beforeObjectCreation(ctx context.Context, obj *unstructured.Unstructured) error {
+	gvk := obj.GroupVersionKind()
+	if gvk == secretGVK {
+		return beforeSecretCreation(ctx, obj)
+	}
+	return nil
+}
+
+func beforeSecretCreation(ctx context.Context, obj *unstructured.Unstructured) error {
+	// If there is any stringData, merge it into data
```
yuwenma, line 102: ➕

yuwenma (commented):
/lgtm
/approve

This PR has so many wonderful pieces and make the code base cleaner and more readable.

PR #384, opened 2024-04-17
mockkubeapiserver: set generation on objects

justinsb:
Aiming for a little bit more fidelity, also enables us to test
observedGeneration (in future).

### DIFF
--- a/mockkubeapiserver/patchresource.go
+++ b/mockkubeapiserver/patchresource.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"reflect"
 	"strconv"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -79,6 +80,11 @@ func (req *patchResource) Run(ctx context.Context, s *MockKubeAPIServer) error {
 		// TODO: Should we treat this like an apply to an empty object?
 
 		patched := body
+
+		if resource.SetsGeneration() {
+			patched.SetGeneration(1)
+		}
+
 		if err := resource.CreateObject(ctx, id, patched); err != nil {
 			return err
 		}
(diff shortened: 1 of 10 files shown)


mockkubeapiserver/patchresource.go
```diff
@@ -124,6 +128,13 @@ func (req *patchResource) Run(ctx context.Context, s *MockKubeAPIServer) error {
 		klog.Infof("skipping write, object not changed")
 		return req.writeResponse(existingObj)
 	} else {
+		bumpGeneration := !reflect.DeepEqual(existingObj.Object["spec"], updated.Object["spec"])
+		if bumpGeneration {
```
tomasaschan, line 132: nit: Intent might be clearer expressed here if we move the negation to the conditional:
```suggestion
		specIsSame := reflect.DeepEqual(existingObj.Object["spec"], updated.Object["spec"])
		if !specIsSame {
```

mockkubeapiserver/putresource.go
```diff
@@ -89,6 +89,13 @@ func (req *putResource) Run(ctx context.Context, s *MockKubeAPIServer) error {
 		return req.writeResponse(original)
 	}
 
+	bumpGeneration := !reflect.DeepEqual(original.Object["spec"], updated.Object["spec"])
+	if bumpGeneration {
```
tomasaschan, line 93: For consistency:
```suggestion
	specIsSame := reflect.DeepEqual(original.Object["spec"], updated.Object["spec"])
	if !specIsSame {
```

mockkubeapiserver/patchresource.go
justinsb: Great suggestion, thank you.  Applying!

mockkubeapiserver/putresource.go
justinsb: Thanks again!

PR #386, opened 2024-04-29
Respect ErrorResults returned from preflight checks

### LINKED ISSUE #374: Error is logged even if result of `status.Preflight` doesn't have one
**What happened**:

I implemented a custom `declarative.Status` like this:
```go
type myStatus {}

func (myStatus) Preflight(ctx context.Context, o declarative.DeclarativeObject) error {
    // omitted: some precondition is not fulfilled
    // want to requeue with backoff and try again after some other controller has had a chance to make the precondition true
    return declarative.ErrorResult{
        Result: result.Result{Requeue: true},
        // note: Err: nil implied here
    }
}

func NewMyStatus() declarative.Status {
    return declarative.StatusBuilder{
        Preflight: myStatus{},
    }
}
```

When my controller reconciles with this, it [logs `"preflight check failed, not reconciling"`](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/pkg/patterns/declarative/reconciler.go#L176) with a stack trace that doesn't point to anywhere in my own code. That's annoying and a little confusing.

**What you expected to happen**: reconciliation is requeued without any errors logged.

Create a declarative controller with a status implementation like the above, reconcile an instance and observe logs.

This seems related to, but is different from #232. I imagine if my status _also_ had a `Reconcile` function, it would be confusing that it gets called even though the preflight check failed (signalled by returning a non-nil result, even if that result has a nil `Err`). That, in turn, happens because [this deferred function](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/blob/master/pkg/patterns/declarative/reconciler.go#L164-L170) doesn't check for the preflight result. (It seems it [once did check](https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/commit/1a4c9a495cb84ddca9b3d5fe60ce3d8720e5924a#diff-91d9000f2817ed2698caf7b294e34ed1bc739bf60b7dc1af3136055b18741312L161-L167), but it's unclear to me why that was removed...)

tomasaschan:
A common pattern in our operators is to use preflight checks to call out to some external system for more information (that can be written e.g. to the status of the `DeclarativeObject`) or to validate that some property of the `spec` meets requirements to be able to expand the channel manifests. If the check fails, sometimes the right behavior is to requeue with backoff - but
often it's not: e.g., if the spec is deemed invalid, there's no reason to requeue the resource. If spec changes we'll get a new event anyway, and until spec changes we won't be able to reconcile.

There is already a `ErrorResult` type that can be used in a Preflight implementation to signal what we want the reconciliation function to return, but it's currently not being respected. This change addresses that.

Fixes #374

### DIFF
--- a/pkg/patterns/declarative/reconciler.go
+++ b/pkg/patterns/declarative/reconciler.go
@@ -95,7 +95,11 @@ type ErrorResult struct {
 }
 
 func (e *ErrorResult) Error() string {
-	return e.Err.Error()
+	if e.Err != nil {
+		return e.Err.Error()
+	}
+
+	return ""
 }
 
 // For mocking
@@ -183,6 +187,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 
 	if r.options.status != nil {
 		if err := r.options.status.Preflight(ctx, instance); err != nil {
+			if errorResult, ok := err.(*ErrorResult); ok {
+				// the user was specific about what they wanted to return; respect that
+				return errorResult.Result, errorResult.Err
+			}
+
 			log.Error(err, "preflight check failed, not reconciling")
 			statusInfo.Err = err
 			return result, statusInfo.Err


justinsb (commented):
Thanks for fixing - I agree this looks like an oversight / partial implementation of ErrorResult.

/approve
/lgtm

PR #388, opened 2024-05-23
fix: use correct path to PodSpec for CronJob

sdowell:
CronJob has a different path to the PodSpec from the other currently supported types of ImageRegistryTransform. This fixes the image transform logic to find the correct path.

**What this PR does / why we need it**: The `ImageRegistryTransform` used for private registries supports a few enumerated types, including `CronJob`. The same `PodSpec` path is used for all types, however `CronJob` has a different path. This change adds special handling for the `CronJob` case.

**Which issue(s) this PR fixes**: N/A

**Special notes for your reviewer**: N/A

**Additional documentation**: N/A


sdowell:
@atoato88 please take a look

PR #389, opened 2024-05-31
Track info.Err in the status field

haiyanmeng:
This allows the reconcilation errors such as errors encountered in the object transformers to be tracked in the field `status.errors`.

Currently, if errors happens during the execution of an object tranformer, the status field is set to:
```
status:
  healthy: false
  phase: InternalError
```

With this CL, the reconciliation errors  are tracked in the field `status.errors`:
```
status:
  errors:
  - 'error building deployment objects: a test error'
  healthy: false
  phase: InternalError
```

### DIFF
--- a/pkg/patterns/addon/pkg/status/kstatus.go
+++ b/pkg/patterns/addon/pkg/status/kstatus.go
@@ -87,7 +87,9 @@ func (k *kstatusAggregator) BuildStatus(ctx context.Context, info *declarative.S
 	}
 
 	shouldComputeHealthFromObjects := info.Manifest != nil && info.LiveObjects != nil
+	currentStatus.Errors = []string{}
 	if info.Err != nil {
+		currentStatus.Errors = []string{info.Err.Error()}
 		switch info.KnownError {
 		case declarative.KnownErrorApplyFailed:
 			currentStatus.Phase = "Applying"


yuwenma:
two questions:
1. I think we normally use the `conditions` for detailed error information. 
2. If tracking an object error, the errors should reflect the current status rather than a error history. Thus, I think we shall have a way to clean up stale errors.

tomasaschan:
I agree this would be better to add to `conditions` rather than `errors`.

Not quite sure how `errors` is used today, though, but I feel like whatever it _is_ used for should probably, eventually, be ported over to be exposed in `conditions` instead. Of course, that should not block you from exposing this info on the resource status, but we might want to be a little strict about adding more things to `errors` if we already know we might want to migrate away from it...

tomasaschan:
Fwiw, this relates to #190.

haiyanmeng:
> Not quite sure how `errors` is used today, though, but I feel like whatever it _is_ used for should probably, eventually, be ported over to be exposed in `conditions` instead.

Do we plan to deprecate the fields `healthy` and `phase` too?

```
type CommonStatus struct {
	Healthy bool     `json:"healthy"`
	Errors  []string `json:"errors,omitempty"`
	Phase   string   `json:"phase,omitempty"`
}
```

tomasaschan:
I don't know if the project has actually made any decisions in this direction, but I think we _should_ 😸 

I base this mostly on [the Kubernetes API Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status) guide, where the following passage can be found:

> Some resources in the v1 API contain fields called phase, and associated message, reason, and other status fields. The pattern of using phase is deprecated. Newer API types should use conditions instead.

But I do think this change makes the behavior better than it currently is, so I'm a little on the fence on whether to just say "that's a project for another day" and go with this as it is...

justinsb (commented):
We should probably look at making this more structured and supporting status.Conditions, but ... LGTM!

/approve
/lgtm

justinsb:
Sorry, missed a lot of the context here when I approved this in VSCode!

I agree that we want to move to status.conditions.  I think that will (likely?) have to be opt-in at the controller level; likely new controllers will stop using the "old" status fields altogether and existing controllers will support both (?).  I think if we do that we can it can be something we can implement relatively quickly, at least for the new case.

I think adding this to the "old" status fields (healthy/phase/errors) makes sense in the interim / for controllers that choose not to support the "new" status fields (conditions)

justinsb:
/test pull-declarative-test

(In the hope that it's just a flake .. we should still fix flakes, but we don't need to fix it in this PR)

haiyanmeng:
> Cannot trigger testing until a trusted user reviews the PR and leaves an /ok-to-test message.

@justinsb , how can I become a trusted user? This would allow me to trigger tests.

PR #393, opened 2024-06-06
Expose more health info as well as counts in Apply results

barney-s:

Expose more health info as well as counts in Apply results
When using applyset as a library we sometimes want more information about the last applied set.
Currently we get boolean (IsHealthy, IsAllApplied).
Having more is useful when consuming this as a library.

### DIFF
--- a/applylib/applyset/results.go
+++ b/applylib/applyset/results.go
@@ -22,6 +22,15 @@ import (
 	"k8s.io/klog/v2"
 )
 
+type ObjectStatus struct {
+	GVK           schema.GroupVersionKind
+	NameNamespace types.NamespacedName
+	IsHealthy     bool
+	IsPruned      bool
+	Message       string
+	Error         error
+}
+
 // ApplyResults contains the results of an Apply operation.
 type ApplyResults struct {
 	total             int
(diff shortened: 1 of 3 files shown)


applylib/applyset/results.go
```diff
@@ -22,70 +22,86 @@ import (
...
-	healthyCount      int
-	unhealthyCount    int
+	Total             int
+	ApplySuccessCount int
+	ApplyFailCount    int
+	PruneSuccessCount int
+	PruneFailCount    int
+	HealthyCount      int
+	UnhealthyCount    int
+	Objects           []ObjectStatus
```
justinsb, line 42: So these were previously unexported so that we could change the implementation and instead offer higher-level methods (AllApplied, All healthy).  What signals do we _want_ to expose?  And which objects appear in Objects?  One answer could be that we every object in the manifest appears in Objects, and then we populate Applied, Is healthy etc at the object level.  But what about objects we are pruning - should they appear in Objects?

I guess the question is ... what do you want to do with this information?  Do we have to commit to exposing everything, or can we make more incremental changes?

barney-s: We mostly want just the objects. Updated the PR to reflect that. Not exposing the counts anymore. Just the list of object status.

justinsb:
Looks great now, thanks @barney-s 

/approve
/lgtm

PR #394, opened 2024-06-07
WIP: test: use go workspace when testing

justinsb:


tomasaschan:
I _think_ this might be what currently blocks builds on a bunch of PRs; if I do this locally (including adding the submodules) the tests that fail on those PRs start passing again.

Is there a reason to not just check in the workspace file in the repo (and use it "always" and not just in the test runs)?

justinsb:
> I think this might be what currently blocks builds on a bunch of PRs; if I do this locally (including adding the submodules) the tests that fail on those PRs start passing again.

So I've re-discovered what I had forgotten, we already do this in `hack/ci/test.sh`, and 

> Is there a reason to not just check in the workspace file in the repo (and use it "always" and not just in the test runs)?

The only reason is that I haven't seen it done elsewhere.  But I agree with you, I'm increasingly wondering when we would _not_ want it.  My biggest concern is that it breaks usage as a library somehow, in the same way that go.mod rewrites can sometimes cause problems.

tomasaschan:
I though the wholesome point of go.work was to enable rewrites and similar tweaks in development while explicitly not affecting usage when importing the module. Maybe worth verifying, but unless I'm completely off here I think it'd be nice to just commit this and ensure the normal workflows with eg plain `go test` Just Works:tm:

PR #395, opened 2024-06-08
Add initial devcontainer

justinsb:


tomasaschan (commented):
Nice!

Not sure how WIP this is, but it might be nice to clear out some of the boilerplate comments (and maybe leave a link to the landing page for devcontaier docs in general at the top). They're quite noisy and don't provide much value beyond "hello world", IMO.

PR #398, opened 2024-06-15
chore: Refactor test helpers into reusable ktest library

justinsb:
This will avoid a circular (module) dependency between
mockkubeapiserver and kdp itself.


justinsb:
Sadly it looks like I can't start using this / remove the old `pkg/test` directories in the same PR.

I feel like I'm using modules "wrong", I guess we should be using a separate repo.  Even if we were using a separate repo, we would have the same sequence of PRs, I guess.

justinsb:
(This is breaking down #383 into its required commits)

hack/ci/test.sh
```diff
@@ -41,6 +41,7 @@ echo "kubectl version is $(kubectl version --client)"
 rm -f go.work go.work.sum
 go work init .
 go work use applylib
+go work use ktest
```
tomasaschan, line 44: This, even more, suggests to me that the workspace file should really be checked into the repo (assuming, which I think is true but haven't double-checked, that all referenced modules live within the same repo). Doing that would, I believe, help with making local development easier, so that one can just clone the repo and have it work both in an editor and with other tooling (e.g. `go test`) without having to first figure out how to make the modules reference each-other in sane ways.

ktest/httprecorder/http_recorder.go
```diff
@@ -0,0 +1,78 @@
...
+	entry.Request = Request{
+		Method: request.Method,
+		URL:    request.URL.String(),
+		Header: request.Header,
+	}
+
+	if request.Body != nil {
+		requestBody, err := io.ReadAll(request.Body)
+		if err != nil {
+			panic("failed to read request body")
```
tomasaschan, line 32: Why not return an error value here, since the signature allows it?

ktest/httprecorder/request_log.go
```diff
@@ -0,0 +1,239 @@
...
+	l.mutex.Lock()
+	defer l.mutex.Unlock()
+
+	for _, entry := range l.entries {
+		entry.Request.Body = resetTimestamp(entry.Request.Body)
+		entry.Response.Body = resetTimestamp(entry.Response.Body)
+	}
+}
+
+func resetTimestamp(body string) string {
```
tomasaschan, line 77: This method replaces `LastTransitionTime` on conditions, but does not care about other timestamps in the payload (e.g. `metadata.creationTimestamp`). Should it?

ktest/httprecorder/request_log.go
```diff
@@ -0,0 +1,239 @@
...
+	b.WriteString("\n")
+	if r.Body != "" {
+		b.WriteString(r.Body)
+		b.WriteString("\n")
+	}
+	return b.String()
+}
+
+type RequestLog struct {
+	// We have seen dropped logs from concurrent requests
```
tomasaschan, line 126:
```suggestion
	// Avoid risking dropped logs from concurrent requests
```
The previous comment makes it a little ambiguous if we've seen dropped logs without this mutex (i.e. the comment explains why we have it) or if we've seen it _despite_ the mutext (implying there might be a concurrency bug somewhere that merits more investigation at some point).

This suggestion is assuming the former is the right interpretation.

ktest/httprecorder/request_log.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+func (l *RequestLog) RegexReplaceURL(find string, replace string) {
+	l.mutex.Lock()
+	defer l.mutex.Unlock()
+
+	for i := range l.entries {
+		request := &l.entries[i].Request
+		u := request.URL
+		r, err := regexp.Compile(find)
```
tomasaschan, line 232: This should probably be done once, outside the loop.

ktest/httprecorder/request_log.go
```diff
@@ -0,0 +1,239 @@
...
+		if err != nil {
+			klog.Fatalf("failed to compile regex %q: %v", find, err)
```
tomasaschan, line 234: Can we make this return the error instead?

hack/ci/test.sh
justinsb: Yes, my only concern is that I haven't seen it elsewhere, but given the pain of not having it checked in you have persuaded me and I think it's worth doing.  If we find something unexpected, then we can revert it.  I'll do a separate PR I guess where we do all this in one (that's the hope, anyway!)

ktest/httprecorder/http_recorder.go
justinsb: Good call - done!

ktest/httprecorder/request_log.go
justinsb: Agreed, done

ktest/httprecorder/request_log.go
justinsb: It makes it harder to use, but I changed this method to accept a `t *testing.T` as its first parameter; this shows it should be used from test code, and then we can easily call t.Fatalf (which is probably equivalent to a panic anyway, but now we're guarding it to be used from test code)

ktest/httprecorder/http_recorder.go
justinsb: Great call.  Done

justinsb:
Cleaned this up, but let's try adding go.work in #383

PR #399, opened 2024-07-03
feat: Enable server-side-apply on direct applier

yuwenma:
This will fix the  "annotation too long" problem when the managed manifests are too large.

Fixes #

**Additional documentation**:

### DIFF
--- a/pkg/patterns/declarative/pkg/applier/direct.go
+++ b/pkg/patterns/declarative/pkg/applier/direct.go
@@ -27,13 +27,20 @@ import (
 )
 
 type DirectApplier struct {
-	inner directApplierSite
+	// Whether to apply the KRM resources using server-side apply. https://kubernetes.io/docs/reference/using-api/server-side-apply/
+	// Note: The server-side apply is stable in Kubernetes v1.22, users should take the responsibility to make sure the cluster
+	// server can support this feature.
+	serverSideApplyPreferred bool
+	inner                    directApplierSite
+}
+
+func (d *DirectApplier) UseServerSideApply() {
+	d.serverSideApplyPreferred = true
 }
 
 var _ Applier = &DirectApplier{}
 
-type directApplier struct {
-}
+type directApplier struct{}
 
 type directApplierSite interface {
 	Run(a *apply.ApplyOptions) error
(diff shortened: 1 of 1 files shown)


pkg/patterns/declarative/pkg/applier/direct.go
```diff
@@ -170,6 +170,7 @@ func (d *DirectApplier) Apply(ctx context.Context, opt ApplierOptions) error {
 		applyOpts.PruneResources = append(applyOpts.PruneResources, r...)
 	}
 
+	applyOpts.ServerSideApply = true
```
tomasaschan, line 173: Should this maybe be behind a feature flag, to allow users to choose when to enable this behavior? If so, could probably be done in the set of flags parsed some 20-ish lines above this one.

(I'm all for it, and would probably enable it first thing, but given that transitioning from CSA to SSA is not always straightforward, it might be worth giving the user a choice here; if nothing else, it enables bumping this library for other reasons without being forced to also manage the transition.)

yuwenma: Thanks for the review @tomasaschan and sorry for the late reply. Changing this to a feature flag makes sense!  I changed it as a feature specific to the direct applier rather than another field in `ApplyOption` (interface level), because we cannot guarantee every implemented applier can support SSA. 

Also, since the SSA is a boolean type, to avoid confusing between unset vs default to false, I used a `preferred` option to distinguish whether it is configured or not.

pkg/patterns/declarative/pkg/applier/direct.go
```diff
@@ -27,13 +27,20 @@ import (
...
 type DirectApplier struct {
-	inner directApplierSite
+	// Whether to apply the KRM resources using server-side apply. https://kubernetes.io/docs/reference/using-api/server-side-apply/
+	// Note: The server-side apply is stable in Kubernetes v1.22, users should take the responsibility to make sure the cluster
+	// server can support this feature.
+	serverSideApplyPreferred bool
+	inner                    directApplierSite
+}
+
+func (d *DirectApplier) UseServerSideApply() {
```
yuwenma, line 37: Use a dedicated function to make the flag more discoverable.

justinsb (approved):

PR #400, opened 2024-07-03
chore: stage ktest module

justinsb:
This should enable us to add a module much more easily, because `go
mod tidy` will work.

### DIFF
--- a/ktest/doc.go
+++ b/ktest/doc.go
@@ -1,0 +1,19 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


justinsb:
This is intended to support #383 , which ends up needing a new shared library, but it proves very difficult to add a module and use it in one commit, even with go workspaces.  I think creating a stub module (as here) should (1) work around the problem and (2) be a reasonable compromise in terms of breaking up commits.

tomasaschan (approved):
/lgtm

tomasaschan:
@justinsb This seems to have already happened on `master`; the changes here are already present (with more stuff added to these files). Could we just close this?

justinsb:
Thanks @tomasaschan  - good point - the bigger PR just merged 🎉 !

/close

PR #402, opened 2024-08-12
Add a callback option to applyset to compute Health of applied objects

barney-s:

We want to provide applyset users a way to plugin their own status compute logic. 
This helps with adding custom status compute for objects not supported by `kstatus` library.


justinsb:
Thanks @barney-s 

/approve
/lgtm

PR #403, opened 2024-08-13
propagate error that occurs during health computation

barney-s:

Capture error as part of health computation and propagate it to health tracker. This will be returned to the caller as part of object result

### DIFF
--- a/applylib/applyset/results.go
+++ b/applylib/applyset/results.go
@@ -22,13 +22,23 @@ import (
 	"k8s.io/klog/v2"
 )
 
+type HealthInfo struct {
+	IsHealthy bool
+	Message   string
+	Error     error
+}
+
+type ApplyInfo struct {
+	IsPruned bool
+	Message  string
+	Error    error
+}
+
 type ObjectStatus struct {
 	GVK           schema.GroupVersionKind
 	NameNamespace types.NamespacedName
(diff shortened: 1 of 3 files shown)


applylib/applyset/results.go
```diff
@@ -113,13 +113,14 @@ func (r *ApplyResults) pruneSuccess(gvk schema.GroupVersionKind, nn types.Namesp
...
 // reportHealth records the health of an object.
-func (r *ApplyResults) reportHealth(gvk schema.GroupVersionKind, nn types.NamespacedName, isHealthy bool, message string) {
+func (r *ApplyResults) reportHealth(gvk schema.GroupVersionKind, nn types.NamespacedName, isHealthy bool, message string, err error) {
 	r.Objects = append(r.Objects, ObjectStatus{
 		GVK:           gvk,
 		NameNamespace: nn,
 		IsHealthy:     isHealthy,
 		IsPruned:      false,
 		Message:       message,
+		Error:         err,
```
justinsb, line 123: This should be something like HealthError I think, so that it's clear that the error applies to health computation.  Maybe we need to replace `IsHealthy bool`  with `Health HealthInfo`, where HealthInfo is something like `struct { Healthy bool; Message string; Error err; }`

justinsb (approved):

PR #405, opened 2024-09-24
chore: update k8s libraries

mclarke47:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Additional documentation**:

PR #406, opened 2024-09-27
chore: Run go work sync to update all modules

tomasaschan:
This fixes the build errors introduced in the merge of #405.

I tried to look at the GH actions config to figure out why this wasn't caught in the review build, but couldn't :/


justinsb:
I _hope_ that #411 was the issue.  I think we should use branches and versioning instead of that pinning anyway.  So hopefully we won't see this problem again (at least in this form), now that you've merged #411 @tomasaschan (thank you!).

PR #408, opened 2025-01-29
docs: document our branching / versioning process

justinsb:
We want to be more systematic about our branching / versioning.

Document the process we were (trying to) follow.  We can also
automate some of the toil here.

### DIFF
--- a/docs/project-management/branching-and-versioning.md
+++ b/docs/project-management/branching-and-versioning.md
@@ -1,0 +1,32 @@
+# Branching and versioning strategy
+
+## Versioning
+
+kubebuilder-declarative-pattern is a library, that works with controller-runtime and kubernetes.
+
+We follow semantic versioning, similar to other tooling in the kubernetes ecosystem.
+
+Because both controller-runtime and client-go introduce changes that require code changes, we follow their versioning.
+Specifically we align with the controller-runtime version, which itself aligns with kubernetes versioning.
+
+Thus:
+
+| kdp version | controller-runtime version | client-go version |
+|---|---|---|
+| v0.20 | v0.20 | v0.32 |
+| v0.19 | v0.19 | v0.31 |
+| v0.18 | v0.18 | v0.30 |
+| ... | ... | ... | 
+| v0.x | v0.x | v0.x+12 |
+
+If we need to release multiple versions of kdp for a single version of controller-runtime (for example if we want to fix a bug),
+we use patch versions for that.
+We want to avoid breaking changes, and we will only make breaking changes on minor version bumps (as far as possible).
+
+## Branches
+
+We maintain a `release-0.x` branch for the kdp major version `0.x`.  Patch versions are cherry-picked to the branch and released.
+
+We cut the release branch with the beta release (i.e. we create `release-1.100` when we tag `1.100.0-beta.1`).
+
+Before beta releases, the master branch is for the next minor version.  If we tag a version, it would be (for example) `1.101.0-alpha.1` (and then `1.101.0-alpha.2` etc), assuming the most recent release branch was `1.100`


docs/project-management/branching-and-versioning.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Branches
+
+We maintain a `release-0.x` branch for the kdp major version `0.x`.  Patch versions are cherry-picked to the branch and released.
+
+We cut the release branch with the beta release (i.e. we create `release-1.100` when we tag `1.100.0-beta.1`).
+
+Before beta releases, the master branch is for the next minor version.  If we tag a version, it would be (for example) `1.101.0-alpha.1` (and then `1.101.0-alpha.2` etc), assuming the most recent release branch was `1.100`
+
+TODO: Should we rename master to main now?
```
tomasaschan, line 33: Yes, let's do it!

docs/project-management/branching-and-versioning.md
```diff
@@ -0,0 +1,35 @@
...
+
+Thus:
+
+| kdp version | controller-runtime version | client-go version |
+|---|---|---|
+| v0.20 | v0.20 | v0.32 |
+| v0.19 | v0.19 | v0.31 |
+| v0.18 | v0.18 | v0.30 |
+| ... | ... | ... | 
+| v0.x | v0.x | v0.x+12 |
```
tomasaschan, line 20: I think we should have the versioning scheme allow for releasing multiple kdp versions per controller-runtime/client-go version; easiest is probably just to increment the patch, but we might want to explicitly call that out here.

docs/project-management/branching-and-versioning.md
```diff
@@ -0,0 +1,35 @@
...
+
+
+TODO: Should we align with client-go version?
```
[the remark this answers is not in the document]
justinsb: Yes, I was just wondering whether controller-runtime might reconsider their decision at some point.  In any case, we just "go up" by 12 versions if we ever change our mind - not a huge problem (in this direction!)

docs/project-management/branching-and-versioning.md
justinsb: I'll try to expand.  My 2c is that while in theory we might want to make breaking changes more often than client-go releases, in practice I don't think that's very likely, and I think that would make the library difficult to consume.  So agree that we should have multiple patch releases - agree that's not clear.  But those patches should avoid breaking changes.

Note that we can still make as many breaking changes as we want on the main branch; we just only release a batch of breaking changes every 4 months or so.  (And of course the ideal number of breaking changes is zero :-) )

docs/project-management/branching-and-versioning.md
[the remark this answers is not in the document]
tomasaschan: I think if we decide to follow controller-runtime and _they_ decide to change their versioning scheme, our decision can stand and we just keep following them :)

docs/project-management/branching-and-versioning.md
justinsb: I added a small paragraph here to describe this - thanks for the suggestion!

docs/project-management/branching-and-versioning.md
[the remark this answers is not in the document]
justinsb: Makes sense - removed the TODO

docs/project-management/branching-and-versioning.md
justinsb: I removed the two TODOs here, and will track them in an issue.  It probably doesn't matter, but I want to think about whether we would tag any releases before or after renaming the branch.

tomasaschan (approved):
/lgtm

PR #413, opened 2025-01-29
chore: script to update golang; update go to 1.23.5

justinsb:
- **chore: create script to update golang**
- **codebot: update go to 1.23.5**


justinsb:
Arguably it doesn't much matter for a library, but I think that suggests we should just stick with latest so that we don't get flagged by scanners.  And testing with latest golang is probably more likely to catch a regression in go (though I don't consider that likely)

tomasaschan (commented):
/lgtm

PR #414, opened 2025-02-11
feat: add observedGeneration to CommonStatus

### LINKED ISSUE #391: KDP Reconciler does not set observedGeneration on the managed resource
**What would you like to be added**: I would like for the KDP reconciler to have built-in support for setting the observedGeneration on the status of the managed resource.

**Why is this needed**: This makes it easier to determine when a KDP controller has reconciled the current spec of the managed resource.

sdowell:
This updates the built in CommonStatus struct to provide observedGeneration behavior when using one of the provided status builders. Custom implementations of the BuildStatus interface could already set observedGeneration if desired, but this provides an opinionated default to the addon implementations.

**What this PR does / why we need it**: Sets `status.observedGeneration` for the built-in `CommonStatus` type. This makes it easier to determine when a KDP controller which uses the Addon object has reconciled the current spec of the managed resource.

Fixes #391

### DIFF
--- a/go.work.sum
+++ b/go.work.sum
@@ -251,6 +251,7 @@ github.com/kr/logfmt v0.0.0-20140226030751-b84e30acd515/go.mod h1:+0opPa2QZZtGFB
 github.com/lithammer/dedent v1.1.0/go.mod h1:jrXYCQtgg0nJiN+StA2KgR7w6CiQNv9Fd/Z9BP0jIOc=
 github.com/lyft/protoc-gen-star/v2 v2.0.3/go.mod h1:amey7yeodaJhXSbf/TlLvWiqQfLOSpEk//mLlc+axEk=
 github.com/matttproud/golang_protobuf_extensions v1.0.2-0.20181231171920-c182affec369 h1:I0XW9+e1XWDxdcEniV4rQAIOPUGDq67JSCiRCgGCZLI=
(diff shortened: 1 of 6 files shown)


tomasaschan:
/ok-to-test

Thanks! I need to wrap my head around what happens for users who already define `ObservedGenerarion` and also use `CommonStatus`, when they upgrade to a version that has this patch - will they get a build error? Runtime error? Or will it just work? - but other than that it looks great!

sdowell:
> /ok-to-test
> 
> Thanks! I need to wrap my head around what happens for users who already define `ObservedGenerarion` and also use `CommonStatus`, when they upgrade to a version that has this patch - will they get a build error? Runtime error? Or will it just work? - but other than that it looks great!

My understanding is that the intended pattern of using the Addon object structs (CommonStatus/CommonSpec) is to embed the addon struct(s) into your CRD struct.

If a user upgrades to the new k-d-p version and already defined an observedGeneration field on their CRD, there would now be duplicate entries of observedGeneration on the struct. For example:
```golang
type Status struct {
	addonv1alpha1.CommonStatus     `json:",inline"`
	// +kubebuilder:default:=0
	ObservedGeneration int64 `json:"observedGeneration"`
}
```
My understanding is that it shouldn't encounter any build errors and the already defined field should take precedence for marshalling/unmarshalling. In this case I think the recommendation would be to replace the already defined field with the CommonStatus builtin to avoid confusion.

And if they didn't already have an observedGeneration field on their struct, it would be added upon upgrading.

Issue #407, opened 2024-11-25
Time to cut a new release?


tomasaschan:
In one of our projects, we have a dependency on `sigs.k8s.io/kubebuilder-declarative-pattern v0.15.0-beta.1.0.20240926141202-cf8082c623b8` where the hash at the end points to the latest commit on `master`. Everything seems to work.

But if we do `go get -u ./... && go mod tidy` in this project, it selects `v0.15.0-beta.2` which then seems to run into a bunch of compatibility issues with other k8s libraries:
```
λ go vet ./...                                                
# sigs.k8s.io/kubebuilder-declarative-pattern/commonclient
../../../pkg/mod/sigs.k8s.io/kubebuilder-declarative-pattern@v0.15.0-beta.2/commonclient/factory_cr15.go:14:28: not enough arguments in call to source.Kind
        have ("sigs.k8s.io/controller-runtime/pkg/cache".Cache, client.Object)
        want ("sigs.k8s.io/controller-runtime/pkg/cache".Cache, object, handler.TypedEventHandler[object, reconcile.Request], ...predicate.TypedPredicate[object])
../../../pkg/mod/sigs.k8s.io/kubebuilder-declarative-pattern@v0.15.0-beta.2/commonclient/restmapper_cr15.go:15:17: undefined: apiutil.NewDiscoveryRESTMapper
# sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/declarative/pkg/applier
../../../pkg/mod/sigs.k8s.io/kubebuilder-declarative-pattern@v0.15.0-beta.2/pkg/patterns/declarative/pkg/applier/direct.go:104:59: f.OpenAPIGetter undefined (type "k8s.io/kubectl/pkg/cmd/util".Factory has no field or method OpenAPIGetter)
../../../pkg/mod/sigs.k8s.io/kubebuilder-declarative-pattern@v0.15.0-beta.2/pkg/patterns/declarative/pkg/applier/direct.go:106:77: too many arguments in call to d.inner.NewFactory(opt).Validator
        have (string, *"k8s.io/cli-runtime/pkg/resource".QueryParamVerifier)
        want (string)
[...]

PR #418, opened 2025-04-25
Update golden test framework

justinsb:
* Move to envtest by default (instead of mock-kubeapiserver)
    
    * Create shared code for capturing requests / normalization.
    
    * Use a more conventional env var WRITE_GOLDEN_OUTPUT
    
    * Support object rewriting


justinsb:
I'm not sure whether anyone is using this framework, I suspect not because it is tied to mock-kubeapiserver.  I think it's easier to just offer envtest by default, and there are a few other simplifications and requirements that brings (e.g. starting the manager).

If we can, I'd also like to change the directory from `tests` to `testsdata` and move to a directory-based format (as we do in KCC), rather than having a bunch of files with the same prefixes in the same directory.

These are breaking, but they are test code and I can't imagine people are using this outside of this repo (based on my experience trying to use it!).  But if anyone knows of any usage, we can always do a v2 or just iterate on this externally.  But because this is breaking (albeit test code) I think we should do it after updating the version.

cc @tomasaschan

tomasaschan:
I'm fine with the breaking change.

Re: updating the version, I need access if you want help :) https://github.com/kubernetes-sigs/kubebuilder-declarative-pattern/issues/407#issuecomment-2720469001

tomasaschan:
/lgtm
/hold

Marking this ready-to-go, but let's hold it until we have #417 done and a new tag available for folks who want a slower migration path (if there are any 🙃). I don't mind this breakage making it into `v0.20`, but agree we should have the `beta.1` tagged without it (we can even tag `beta.2` with it, if we want to have a prerelease that includes it).

justinsb:
A massive diff, the majority of which is because I switched from json.Marshal to json.MarshalIndent.  I could easily split that out into a separate PR though ... or just not do it at all.  There will be changes to most golden-test-data lines (because of UID and resourceVersion simplification) which is why I thought it might be worth doing it in this PR.

dev/test
```diff
@@ -27,6 +27,9 @@ cd "${REPO_ROOT}"
 
 set -x
 
+# Download the kubebuilder assets for envtest
+export KUBEBUILDER_ASSETS=$(go run sigs.k8s.io/controller-runtime/tools/setup-envtest@latest use -p path)
```
tomasaschan, line 31: As of https://github.com/kubernetes-sigs/controller-runtime/pull/3135 this isn't really strictly necessary anymore. Would be worthwhile getting rid of while we're doing an overhaul, imo!

ktest/testharness/kubeapiserver.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+		s.restConfig = restConfig
+		t.Cleanup(func() {
+			if err := env.Stop(); err != nil {
+				t.Errorf("error stopping envtest: %v", err)
+			}
+		})
+	} else {
+		// Removed for now...
+		t.Fatalf("mockkubeapiserver is not supported with this test harness")
```
tomasaschan, line 45: Are you planning to add this back? If not, I suggest making the parameter non-nil and just getting rid of the branch here.

examples/guestbook-operator/controllers/tests/patches-stable.in.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: addons.example.org/v1alpha1
 kind: Guestbook
 metadata:
   name: guestbook-sample
+  namespace: default
```
tomasaschan, line 5: Is there something in the new test harness that requires this to be specified with a default value, or are you just doing that to be more explicit? Since the actual API server does not require it I think it would be valuable if we also didn't in the test harness.

dev/test
justinsb: OK that is very cool.  I don't think it is released yet on a controller-runtime tag (at least not that I could see).  We could keep our main/master branch close to the upstream controller-runtime main/master branch, now that we have branches!

examples/guestbook-operator/controllers/tests/patches-stable.in.yaml
justinsb: So with envtest we need to create the objects for real, and kube-apiserver rejects the namespace-scoped object without a namespace.   (We're not using kubectl, which would apply the default namespace for us)

We _might_ be able to get away with not creating the object, but where there are "side objects" / dependencies those do need to created.  I personally see this as the test suite now catching an invalid object.

ktest/testharness/kubeapiserver.go
justinsb: I added it back.  The output is slightly different because of the api discovery v2 features added in k/k, but I figure we can tackle that separately (if we choose to)

tomasaschan (commented):
/lgtm

Great work on this!

ktest/httprecorder/kube_normalize.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	if found {
+		unstructured.SetNestedField(u.Object, replace, path...)
+	}
+}
+
+// rewriteAPIV3Hashes removes some semi-random hashes from the api discovery response.
+// It is semi-random because these are the hashes of all the types in that group,
+// so should be stable for a single k8s patch release.
+func rewriteAPIV3Hashes(t *testing.T, body map[string]any) {
```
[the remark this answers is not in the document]
justinsb: Hmmm good question.  So I did it because the hashes appear in the URL and they were appearing (and changing on patch kube version bumps).  I think for discovery requests we only replace the body, but let me check.

(I think it's important that the discovery requests have _a_ footprint because one of the problems we had in the past was that we were doing discovery every time.  That's less important now that discovery is much more efficient, but I want to avoid regression there and on similar inefficiencies that don't change the outcome)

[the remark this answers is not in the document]
justinsb: You were spot on.  I added openapi/v3 to the list of endpoints which should be removed for length, and that let me remove this function entirely.  There are a few remaining discovery queries with hashes in the URL, but query parameters are much easier to replace so I did that.

justinsb:
Personally I think we should merge this, because I think it's only going to touch test code and not the most widely used test code at that...  If anything we want the envtest support because without the tests will use mockkubeapiserver, which does not accurately reflect modern kube.  (I would like to fix that too, but not in this PR!)

tomasaschan (approved):
Sweet!

/lgtm

examples/guestbook-operator/controllers/tests/patches-stable.in.yaml
tomasaschan: I finally got around to bumping this library in one of our internal projects, and it turns out this has another potentially unintended - and for us definitely disruptive - consequence.

Some context: our operator is configured with a `declarative.WithStatus()` node that includes a preflight request; that preflight request fetches some information we need using a remote service call, and stores that on the resource's `status` _including calling `client.Status().Update()` to persist the info_, before it gets to the templating step. This ensures that the template can read from both `spec` and `status` and have all the info it needs to expand.

Now, in the "new" validator function, because the YAML read from the input file is actually sent to the cluster and cast to `runtime.Object`, and because `status` is a subresource on our CRD, this means the status information in `input.yaml` is not present when templating runs. This means we can't verify anything about the parts of the output that depend on information from `status`. Unless there's a hook somewhere to re-inject that information?

PR #419, opened 2025-04-26
chore: Add automatic tag and branch creation, triggered by version/TAG_VERSION

justinsb:
Doing this makes permissions easier (no need for admin access) and also
ensures that tag operations go through PR review (which is nice from
a compliance perspective).

Inspired by similar work done by the kOps maintainers.


justinsb:
This doesn't replace the need to get you those permissions @tomasaschan , but hopefully it makes it less urgent!

PR #421, opened 2025-04-28
chore: update version of kubectl used in tests

justinsb:
We want to use newer kubectl versions in general.

This version is also helpful for golden tests because it prefers json over protobuf.

### DIFF
--- a/hack/ci/test.sh
+++ b/hack/ci/test.sh
@@ -31,7 +31,7 @@ REPO_ROOT=$(pwd)
 if [[ ! -f "bin/kubectl" ]]; then
   echo "Downloading kubectl to bin/kubectl"
   mkdir -p bin/
-  curl -L -o bin/kubectl https://dl.k8s.io/release/v1.27.0/bin/linux/amd64/kubectl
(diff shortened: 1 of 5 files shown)


hack/ci/test.sh
```diff
@@ -31,7 +31,7 @@ REPO_ROOT=$(pwd)
 if [[ ! -f "bin/kubectl" ]]; then
   echo "Downloading kubectl to bin/kubectl"
   mkdir -p bin/
-  curl -L -o bin/kubectl https://dl.k8s.io/release/v1.27.0/bin/linux/amd64/kubectl
+  curl -L -o bin/kubectl https://dl.k8s.io/release/v1.32.1/bin/linux/amd64/kubectl
```
tomasaschan, line 34: Ah, this (or rather not doing this) is why my attempts at this in #417 kept failing.

tomasaschan (commented):
/lgtm but we might bundle this in #417 since I guess it makes sense to use a version of kubectl that matches the version of client-go.

justinsb:
Definitely in favor of bundling these into fewer PRs (and #417 in particular).  I split them up usually so that I can figure out what fails if we only do half of a bigger PR :-)

PR #423, opened 2026-02-12
Work around branch protection

tomasaschan:
This is just to work around the fact that git won't let me push to my feature branch :/


tomasaschan:
Superceded by #424

PR #426, opened 2026-04-03
Move per-reconcile and per-resource log messages to V(2)

chmill-zz:
Several `Info()` log calls emit on every reconcile loop iteration or for every watched resource, even when nothing has changed. At default verbosity these dominate log output and obscure operationally significant messages.

This PR moves these messages from `Info()` (V(0)) to `V(2).Info()`, keeping them available for debugging while reducing noise at default verbosity.

### Messages moved to V(2)

| Message | File | Why |
|---|---|---|
| `Got status of resource:` | `pkg/patterns/addon/pkg/status/kstatus.go` | Logged for every resource on every status check, including no-op "Current" status |
| `watch began` / `watch closed` | `pkg/patterns/declarative/pkg/watch/dynamic.go` | Watch lifecycle bookkeeping |
| `broadcasting event` | `pkg/patterns/declarative/pkg/watch/dynamic.go` | Every individual watch event |
| `reconciling` | `pkg/patterns/declarative/reconciler.go` | Every reconcile loop entry |
| `built deployment objects` | `pkg/patterns/declarative/reconciler.go` | Every reconcile |
| `setting namespace` | `pkg/patterns/declarative/reconciler.go` | Every reconcile |
| `injecting owner references` | `pkg/patterns/declarative/reconciler.go` | Every reconcile |
| `loading package` | `pkg/patterns/addon/pkg/loaders/*.go` | Every reconcile (3 loader implementations) |
| `using specified version` | `pkg/patterns/addon/pkg/loaders/fs.go` | Every reconcile |

### Rationale

The logr library convention is:
- **V(0) / Info()** — operational: state transitions, errors, decisions
- **V(1)** — useful debugging detail
- **V(2)+** — per-iteration, per-resource trace output

These messages are all per-iteration trace output. They don't indicate state changes or problems — they confirm the system is doing its normal work. Operators who need trace-level detail can run with `-v=2`.

### No behavior change

This only affects log output at default verbosity. All messages remain available at V(2). No functional behavior is changed.

### DIFF
--- a/pkg/patterns/addon/pkg/loaders/fs.go
+++ b/pkg/patterns/addon/pkg/loaders/fs.go
@@ -105,7 +105,7 @@ func (c *ManifestLoader) ResolveManifest(ctx context.Context, object runtime.Obj
 
 		log.WithValues("channel", channelName).WithValues("version", id).Info("resolved version from channel")
 	} else {
-		log.WithValues("version", version).Info("using specified version")
+		log.WithValues("version", version).V(2).Info("using specified version")
 	}
 	s := make(map[string]string)
 	s, err = c.repo.LoadManifest(ctx, componentName, id)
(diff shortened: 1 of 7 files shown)


tomasaschan:
/lgtm
/approve 

Thanks!

